PR #6, opened 2020-08-17
Captcha field

IvanGrginovInf:
Changelog
---
* Added accessible basic math captcha field
* Added captcha check to Dynamics CRM route (will probably abstract that to the parent Route after adding more routes)

### DIFF
--- a/src/routes/class-dynamics-crm-route.php
+++ b/src/routes/class-dynamics-crm-route.php
@@ -15,6 +15,7 @@ namespace Eightshift_Forms\Rest;
 use Eightshift_Forms\Core\Filters;
 use Eightshift_Forms\Integrations\Dynamics_CRM;
 use Eightshift_Libs\Core\Config_Data;
+use Eightshift_Forms\Captcha\Basic_Captcha;
 
 /**
  * Class Dynamics_Crm_Route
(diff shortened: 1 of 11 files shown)


src/routes/class-dynamics-crm-route.php
```diff
@@ -96,6 +102,11 @@ public function route_callback( \WP_REST_Request $request ) {
    */
   protected function defined_responses(string $response_key, array $data = []): array {
     $responses = [
+      'wrong-captcha' => [
+        'code' => 401,
```
dingo-d, line 106: I don't think 401 is a good status code. 401 is unauthorized, and Recaptcha doesn't have anything to do with authorization. When in doubt 404 it out 😄

IvanGrginovInf: True although I'm not too happy with 404 being default (since "not found" is also technically not true). Will leave it as 404 tho 👍

dingo-d: I've searched a bit and found that you could use 429, even though that's too many requests (used when doing rate-limiting), so it kinda felt odd to be used for captcha 🤷‍♂️  but maybe it could be used.

IvanGrginovInf: Yeah a bit better (much better than 401) but not 100% true as well. @dingo-d how about just `403`? [Source](https://en.wikipedia.org/wiki/HTTP_403):

> HTTP 403 is a HTTP status code meaning access to the requested resource is forbidden for some reason
> ...
> The server understood the request, but is refusing to authorise it.

We're not using any authorization but I also found this [here](https://httpstatuses.com/403):

> ...However, a request might be forbidden for reasons unrelated to the credentials.

dingo-d: 403 is unauthenticated, and again, the captcha isn't used for authentication purposes. 401 is when you are authenticated but doesn't have the right permissions, and 403 is when you want to access a part of the web (app) but you didn't log in. 

I'm ok it being 429, as bots could try multiple times 🤷‍♂️ 

I'm eager to see the results when this is implemented if they'll have spam or not 😄

PR #8, opened 2020-08-26
Email route

IvanGrginovInf:
Changelog
---
* Added ability to send emails via form
* Better UX for the select block
* Added inline (in editor) editable labels
* Added ability to register form blocks inside your project (filter list of available blocks used in forms)
* Prevented some unwanted interaction with form elements inside editor (such as accidentally submitting the form by clicking the submit button while you're editing it)
* Separated all form options inside tabs (like we have for responsive wrapper options)

### DIFF
--- a/src/blocks/custom/form/components/form-email-options.js
+++ b/src/blocks/custom/form/components/form-email-options.js
@@ -1,18 +1,67 @@
 import { __ } from '@wordpress/i18n';
 import { Fragment } from '@wordpress/element';
-import { SelectControl } from '@wordpress/components';
+import { TextControl, BaseControl } from '@wordpress/components';
+import { RichText } from '@wordpress/block-editor';
 
 export const FormEmailOptions = (props) => {
   const {
-    type,
-    crmEntitiesAsOptions,
-    dynamicsEntity,
-    isDynamicsCrmUsed,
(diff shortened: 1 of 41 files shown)


iruzevic (approved):

src/blocks/custom/form/components/form-email-options.js
```diff
@@ -0,0 +1,67 @@
...
+  } = props;
+
+  const defaultMessage = message ?? 'Message from user [[name]]:<br><br>[[message]]';
+
+  return (
+    <Fragment>
+      {onChangeTo &&
+        <TextControl
+          label={__('To', 'eightshift-forms')}
+          help={__('Email address to which you wish to receive emails.', 'eightshift-forms')}
```
gabriel-glo, line 26: Shouldn't this say 'send emails'?

gabriel-glo (approved):
Just one small comment to look over

src/blocks/custom/form/components/form-email-options.js
IvanGrginovInf: Hmm good point, it's more clear, will change. Although both are technically true (because the person setting up the forms is likely the person receiving them). 😄

iruzevic (approved):

PR #9, opened 2020-09-08
More improvements

IvanGrginovInf:
Changelog
---
* Added ability to pre-fill some fields. You can now set data sources in your project which are then available for prefilling in the form fields (currently the option exists only on `select` and `radio` blocks). For example you can have a `<select>` that's pre-filled from a 3rd party API source.
* Added ability to set placeholders inside form field's `value` (for example `[[email]]`). These will be replaced with values of other form fields before sending. This allows you (for example) to have 2 fields in a form that are concatenated and sent as 1 field. See [wiki for more info](https://github.com/infinum/eightshift-forms/wiki/Sending-multiple-form-fields-as-a-single-field.).
* Added **Prevent sending** option to fields (if enabled that field won't be sent to the processing route on form submission - this is useful for CRM route where you will get an error if you send extra fields not recognised by CRM)
* Added data-attribute changes when form is submitted (so "form successfully submitted" event can be caught in your project using `MutationObserver`)
* Fixed `radio-item` markup to allow for custom radio buttons (same as checkbox) while still being accessible
* Added Style variation for radio block (list vs flex)
* Added Dynamics CRM route for reading entities from CRM
* Added Transient Cache object
* Some bugfixes


composer.json
```diff
@@ -26,8 +26,8 @@
   },
   "require-dev": {
     "dealerdirect/phpcodesniffer-composer-installer": "^0.7.0",
-    "giacocorsiglia/wordpress-stubs": "^5.1",
-    "infinum/coding-standards-wp": "0.4.3"
+    "infinum/coding-standards-wp": "0.4.3",
+    "php-stubs/wordpress-stubs": "^5.5"
```
dingo-d, line 30: Do you have tests? What do you need stubs for?

src/blocks/custom/checkbox/checkbox.php
```diff
@@ -9,17 +9,18 @@
...
+$name            = $attributes['name'] ?? '';
+$value           = $attributes['value'] ?? '';
+$label           = $attributes['label'] ?? '';
+$checkbox_id     = $attributes['id'] ?? '';
+$classes         = $attributes['classes'] ?? '';
+$theme           = $attributes['theme'] ?? '';
+$is_checked      = isset( $attributes['isChecked'] ) && $attributes['isChecked'] ? 'checked' : '';
+$is_disabled     = isset( $attributes['isDisabled'] ) && $attributes['isDisabled'] ? 'disabled' : '';
+$is_read_only    = isset( $attributes['isReadOnly'] ) && $attributes['isReadOnly'] ? 'readonly' : '';
+$is_required     = isset( $attributes['isRequired'] ) && $attributes['isRequired'] ? 'required' : '';
```
dingo-d, line 22: Could WP functions be used here?

https://developer.wordpress.org/reference/functions/checked/
https://developer.wordpress.org/reference/functions/selected/
https://developer.wordpress.org/reference/functions/disabled/
https://developer.wordpress.org/reference/functions/readonly/

?

It could be overkill but could be cool to use 🤷‍♂️  Thoughts?

src/blocks/custom/checkbox/components/checkbox-options.js
```diff
@@ -90,6 +92,15 @@ export const CheckboxOptions = (props) => {
           onChange={onChangeIsRequired}
         />
       }
+
+      {onChangePreventSending &&
+        <ToggleControl
+          label={__('Do not send?', 'eightshift-forms')}
+          help={__('If enabled this field will not be sent when form is submitted.', 'eightshift-forms')}
```
dingo-d, line 99:
```suggestion
          help={__('If enabled, this field won\'t be sent when the form is submitted.', 'eightshift-forms')}
```

src/blocks/custom/input/components/input-options.js
```diff
@@ -146,6 +148,15 @@ export const InputOptions = (props) => {
           onChange={onChangeIsRequired}
         />
       }
+
+      {onChangePreventSending &&
+        <ToggleControl
+          label={__('Do not send?', 'eightshift-forms')}
+          help={__('If enabled this field will not be sent when form is submitted.', 'eightshift-forms')}
```
dingo-d, line 155:
```suggestion
          help={__('If enabled, this field won\'t be sent when the form is submitted.', 'eightshift-forms')}
```

src/blocks/custom/select/components/select-options.js
```diff
@@ -32,6 +66,15 @@ export const SelectOptions = (props) => {
         />
       }
 
+      {onChangePreventSending &&
+        <ToggleControl
+          label={__('Do not send?', 'eightshift-forms')}
+          help={__('If enabled this field will not be sent when form is submitted.', 'eightshift-forms')}
```
dingo-d, line 72:
```suggestion
          help={__('If enabled, this field won\'t be sent when the form is submitted.', 'eightshift-forms')}
```

src/blocks/custom/textarea/components/textarea-options.js
```diff
@@ -100,6 +102,15 @@ export const TextareaOptions = (props) => {
           onChange={onChangeIsReadOnly}
         />
       }
+
+      {onChangePreventSending &&
+        <ToggleControl
+          label={__('Do not send?', 'eightshift-forms')}
+          help={__('If enabled this field will not be sent when form is submitted.', 'eightshift-forms')}
```
dingo-d, line 109:
```suggestion
          help={__('If enabled, this field won\'t be sent when the form is submitted.', 'eightshift-forms')}
```

dingo-d (changes requested):
Left some minor comments

src/blocks/custom/checkbox/checkbox.php
gabriel-glo: I don't think they can. As it states they output the attributes and should be used inside form elements.
These variable should contain boolean values and used as arguments to these function listed above.

src/blocks/custom/checkbox/checkbox.php
```diff
@@ -39,9 +40,11 @@ class="<?php echo esc_attr( "{$classes} {$block_class}__checkbox js-{$block_clas
       <?php echo esc_attr( $is_disabled ); ?>
```
gabriel-glo, line 40: You could however, use the functions from Denis' comment here, where you output the attributes

gabriel-glo (commented):
Think about if you want to use these form element functions for checking, selecting or disabling form elements. Function that is missing is the one that will handle required form elements.

composer.json
IvanGrginovInf: For when developing the plugin, I open the root of the plugin (because it's installed in the actual project) and in that case my editor doesn't know about any WordPress function and phpcs goes crazy. These stubs fix that issue.

dingo-d: Hmmm, haven't thought about that tbh. That is actually really smart! I always open my projects in the root of WP so that PHPStorm can autocomplete everything.

Do you have documentation about WP specific functions with stubs? Like when you hover over them?

src/blocks/custom/checkbox/checkbox.php
dingo-d: The `$is_checked` variable does a boolean check with `isset` and what `$attributes['isChecked']` inside the ternary. The point is that you'd just have the booleans in these variables, and down in the code you'd use the function with the boolean provided.

Looking at it now, I think it might be an overkill.

src/blocks/custom/checkbox/checkbox.php
dingo-d: Yeah, that was my intention, but I don't think that we gain much (other than following the 'WordPress way').

iruzevic (approved):
Looks good. I will approve. Please address all issues from Denis and GG

composer.json
IvanGrginovInf: > Do you have documentation about WP specific functions with stubs? Like when you hover over them?

Yup, everything is there, description, parameter type hints / description, etc (all except the actual body of function haha)

> I always open my projects in the root of WP so that PHPStorm can autocomplete everything.

Usually me too but here in the root I have a different repository (project I'm working on) and my Gitlens doesn't really know how to handle those 2 repos meaning I have to do it like this (have 2 editor window for project, another for the plugin).

src/blocks/custom/checkbox/checkbox.php
IvanGrginovInf: Yeah in attributes you do have a boolean so we could in theory use this.

I personally find them confusing a bit, especially in this situation ("checking checked vs current" what is checked vs what is current in this case?).

Looking at the docs / example, it seems it makes most sense when you're in a form's controller (you have the current saved value vs what you received in the form. Then you compare the two to figure out of checkbox should be checked or not). 

In our case, we only really have 1 value (the value that is passed to the component) so I think it makes more sense how it is currently.

IvanGrginovInf:
> Think about if you want to use these form element functions for checking, selecting or disabling form elements. Function that is missing is the one that will handle required form elements.

@gabriel-glo really not sure what you mean by this, could you please clarify?

src/blocks/custom/checkbox/checkbox.php
IvanGrginovInf: See my comment from https://github.com/infinum/eightshift-forms/pull/9#discussion_r485419154

I mean the equivalent would be (which is a bit ugly IMO):

```php
$is_checked = checked( $attributes['isChecked'] ?? false, true, false );

// Render / markup starts

<?php echo esc_attr( $is_checked ); ?>
```

If we did just use `checked` on output it would be cleaner BUT then we're dealing with attributes inside the view which I'm not a big fan off (and no other block of ours does that as far as I know):

```php

// Render / markup starts

<?php checked( $attributes['isChecked'] ?? false ); ?>

**Comment by gabriel-glo** [comment]
I was referring to these functions:
https://developer.wordpress.org/reference/functions/checked/
https://developer.wordpress.org/reference/functions/selected/
https://developer.wordpress.org/reference/functions/disabled/
https://developer.wordpress.org/reference/functions/readonly/

and if you should use them in form components, as they are more aligned to the WordPress way. You already have a good way of determining and outputting these values. If you choose to use these functions you will still need to use your implementation for outputting the `required` attribute.

**Comment by IvanGrginovInf** [comment]
> I was referring to these functions:
> https://developer.wordpress.org/reference/functions/checked/
> https://developer.wordpress.org/reference/functions/selected/
> https://developer.wordpress.org/reference/functions/disabled/
> https://developer.wordpress.org/reference/functions/readonly/
> 
> and if you should use them in form components, as they are more aligned to the WordPress way. You already have a good way of determining and outputting these values. If you choose to use these functions you will still need to use your implementation for outputting the `required` attribute.

Yeah agree, then we'd have less consitency. Same thing with some custom data attributes that are handled in the same way (toggle in editor - such as `data-prevent-sending`).

**Review by dingo-d — APPROVED** [review]

**Review by gabriel-glo — APPROVED** [review]

**Review by iruzevic — APPROVED** [review]

File: src/blocks/custom/select/select.php [hunk 8]
```diff
@@ -18,12 +18,26 @@
 $theme           = $attributes['theme'] ?? '';
 $should_prefill  = $attributes['prefillData'] ?? false;
 $prefill_source  = $attributes['prefillDataSource'] ?? '';
+$hide_loading    = $attributes['hideLoading'] ?? true;
 $is_disabled     = isset( $attributes['isDisabled'] ) && $attributes['isDisabled'] ? 'disabled' : '';
 $prevent_sending = isset( $attributes['preventSending'] ) && $attributes['preventSending'] ? 'data-do-not-send' : '';
 
 $block_classes = Components::classnames([
   $block_class,
+  "js-$block_class",
```
dingo-d, line 27:
```suggestion
  "js-{$block_class}",
```

Readability

dingo-d (changes requested):
One minor comment left

src/blocks/custom/basic-captcha/components/basic-captcha-editor.js
```diff
@@ -8,6 +8,8 @@ export const BasicCaptchaEditor = (props) => {
     },
   } = props;
 
+  console.log(attributes);
```
volfkarlo, line 11: This is intended or?

volfkarlo (approved):

src/blocks/custom/basic-captcha/components/basic-captcha-editor.js
dingo-d, line 11: console statement

dingo-d (commented):
Minor comment

src/blocks/custom/basic-captcha/components/basic-captcha-editor.js
IvanGrginovInf: Good catch, fixed.

IvanGrginovInf: Good catch, fixed.

dingo-d (approved):

PR #11, opened 2020-10-13
Buckaroo integration

IvanGrginovInf:
Changelog
---
* Added buckaroo iDEAL route
* Added buckaroo Emandate route
* Added Buckaroo response handler route (allows custom actions to be run after user was redirected back from Buckaroo)
* Added HMAC authorization class (this is used to allow routes to require an authorization hash for access - easy route only needs to provide secret salt)
* Added automated unit tests
* Refactored / abstracted a lot of route code


src/integrations/buckaroo/class-buckaroo.php
```diff
@@ -352,6 +357,8 @@ protected function build_post_body_for_emandate( string $debtorreference, string
 
     $post_array['Services']['ServiceList'][] = $service_array;
 
+    error_log(print_r($post_array, true));
```
iruzevic, line 360: a rouge error log here

iruzevic (changes requested):

src/exception/missing-filter-info-exception.php
```diff
@@ -0,0 +1,40 @@
...
+   *                              not readable.
+   * @param \Exception $exception Exception that was thrown by the view file.
+   *
+   * @return static
+   *
+   * @since 0.1.0
+   */
+  public static function view_exception( $filter, $key ) {
+    $message = sprintf(
+      \esc_html__( 'Missing a required key %s in %s filter, please provide that as part of return array on that filter', 'eightshift-libs' ),
```
iruzevic, line 33: wrong textdomain

src/routes/class-send-email-route.php
```diff
@@ -78,6 +74,7 @@ public function route_callback( \WP_REST_Request $request ) {
     return \rest_ensure_response([
       'code' => 200,
       'message' => esc_html__( 'Email sent', 'd66' ),
```
iruzevic, line 76: wrong textdomain

tests/BaseTest.php
```diff
@@ -0,0 +1,62 @@
+<?php namespace EightshiftFormsTests;
+
+use \Brain\Monkey;
+use \Brain\Monkey\Functions;
+
+class BaseTest extends \Codeception\Test\Unit
+{
+
+  const WP_REDIRECT_ACTION = 'eightshift_forms_test/wp_safe_redirect_happened';
+  const WP_MAIL_ACTION = 'eightshift_forms_test/wp_mail_happened';
```
iruzevic, line 10: is this eightshift_forms_test correct?

iruzevic (changes requested):
a small changes

tests/BaseTest.php
IvanGrginovInf: Yup, these are actions **specifically for unit tests** which we can use to see if some native WP function happened that would be otherwise hard to test (such as `wp_mail()` and `wp_safe_redirect()` in this case) .

composer.json
```diff
@@ -1,54 +1,61 @@
...
-  "homepage": "https://eightshift.com/",
-  "license": "MIT",
-  "authors": [
-    {
-      "name": "Eightshift team",
-      "email": "team@eightshift.com",
-      "homepage": "https://eightshift.com/",
-      "role": "Developer / IT Manager"
+    "name": "eightshift/eightshift-forms-plugin",
+    "version": "1.2.2",
```
dingo-d, line 3: I think this gets picked up from tags, so you don't need to add it.

composer.json
```diff
@@ -1,54 +1,61 @@
...
+    "description": "This repository contains all the tools you need to start building a modern WordPress project.",
```
dingo-d, line 4: This should probably be changed

composer.json
```diff
@@ -1,54 +1,61 @@
...
+    "minimum-stability": "dev",
+    "prefer-stable": true,
+    "config": {
+        "sort-packages": true,
+        "optimize-autoloader": true,
+        "process-timeout": 2000
+    },
+    "scripts": {
+        "check-cs": "@php ./vendor/squizlabs/php_codesniffer/bin/phpcs --runtime-set testVersion 7.1-",
+        "fix-cs": "@php ./vendor/squizlabs/php_codesniffer/bin/phpcbf --runtime-set testVersion 7.1-"
```
dingo-d, line 59: These should be set from the `phpcs.xml.dist`. Also, since you have tests, it's good to add additional scripts here as well 🙂 

I think I've added some in the testing with Codeception chapter in the handbook.

src/blocks/custom/form/components/form-buckaroo-options.js
```diff
@@ -0,0 +1,81 @@
...
+        />
+      }
+
+      {onChangeEmandateDescription && service === 'emandate' &&
+        <TextareaControl
+          label={__('Emandate description', 'eightshift-forms')}
+          value={emandateDescription}
+          help={__('A description of the (purpose) of the emandate. This will be shown in the emandate information of the customers\' bank account. Max 70 characters.', 'eightshift-forms')}
+          onChange={(newValue) => {
+            const MAX_CHARS_IN_FIELD = 70;
```
dingo-d, line 45: This constant must be here? Can't it be defined outside the `return` statement, and used here?

src/blocks/custom/form/form.php
```diff
@@ -45,8 +53,9 @@ class="<?php echo esc_attr( "{$block_class}__form js-{$block_class}-form" ); ?>"
     action="<?php echo esc_attr( $form_action ); ?>"
     method="<?php echo esc_attr( $form_method ); ?>"
     target="<?php echo esc_attr( $form_target ); ?>"
-    <?php ! empty( $form_id ) ? printf( 'id="%s"', esc_attr( $form_id ) ) : ''; ?>
     data-form-type="<?php echo esc_attr( $form_type ); ?>"
+    <?php ! empty( $form_id ) ? printf( 'id="%s"', esc_attr( $form_id ) ) : ''; ?>
```
dingo-d, line 57: What is the type of `$form_id`? If it's integer the `%d` should be used. With that, technically, you don't need to escape it when passing to `printf` as an argument (cool trick 😄 ).

Or you can 'escape' it using `(int)` typecasting. Which is a valid way of escaping integers.

composer.json
IvanGrginovInf: Good idea, added just the unit one for now + phpstan.

src/class-actions.php
```diff
@@ -0,0 +1,27 @@
...
+  /**
+   * This action should be used when doing Buckaroo integration. It will run after the user
+   * has been redirected from Buckaroo (with appropriate status - success, cancel, error, reject)
+   * back to our site but before he is redirected to the success / error page defined in Form's options.
+   *
+   * The action's callback needs to be provided in your project.
+   *
+   * @var string
+   */
+  const BUCKAROO_RESPONSE_HANDLER = 'eightshift_forms/buckaroo_response_handler';
```
dingo-d, line 26: So the only usage of this class is to define the constant?  👀 

You could have used interface for it, but still, I'm failing to see the reasoning behind extracting this to a class. Unless it's reused in 10 different places, in which case interface makes a bit more sense (to me).

src/class-filters.php
```diff
@@ -13,9 +13,50 @@
  * The Filters class, used for defining available filters.
  */
 class Filters {
-  const DYNAMICS_CRM           = 'eightshift_forms/dynamics_info';
-  const ALLOWED_BLOCKS         = 'eightshift_forms/allowed_blocks';
-  const GENERAL                = 'eightshift_forms/general_info';
-  const PREFILL_GENERIC_MULTI  = 'eightshift_forms/prefill/multi';
-  const PREFILL_GENERIC_SINGLE = 'eightshift_forms/prefill/single';
+
```
dingo-d, line 16: The same is as with actions. You could maybe have interface `Hooks` if this is being reused across the codebase.

src/class-main.php
```diff
@@ -38,6 +37,13 @@
  */
 class Main extends Lib_Core {
 
+  /**
+   * Set this to true if you need dependency injection in tests.
+   *
+   * @var boolean
+   */
+  protected $is_test = false;
```
dingo-d, line 45: If this is used only in this class, it can be set to `private`. I generally prefer using either defined constants or env variables for this. Env variables are great because you can override them in tests (if using wp-browser).

src/class-main.php
```diff
@@ -115,4 +151,68 @@ protected function get_service_classes() : array {
...
+
+  /**
+   * Get the list of services to register.
+   *
+   * A list of classes which contain hooks.
+   *
+   * @return array<string> Array of fully qualified class names.
+   */
+  protected function get_service_classes() : array {
+    return $this->is_test ? $this->test_service_classes() : $this->prod_service_classes();
```
dingo-d, line 206: How do you set this during tests, especially if it's protected?

src/blocks/custom/form/form.php
IvanGrginovInf: ID as in `id, class`, not `id` as in post type id 😄  (meaning it's a string, not int)

src/class-main.php
```diff
@@ -115,4 +151,68 @@ protected function get_service_classes() : array {
...
+  }
+
+  /**
+   * Provides additional / different services depending on if we're in test or not.
+   *
+   * @param  boolean $is_test Set to true if running tests.
+   * @return void
+   */
+  public function set_test( bool $is_test ): void {
```
dingo-d, line 215: This is interesting. So during the test, you need to first set this to `true`, then you can bootstrap the plugin? A bit odd approach, but if it works 🤷‍♂️

src/class-actions.php
IvanGrginovInf: There is another class for `filters` (`class-filters.php`) which has ~9 filters. This one is for `actions`. Coincidentally there's only 1 action right now but this will probably change in the future.

It serves as a simple list of available filters / actions - both for future documentation as well as just you know, to keep all these constants in a single place. I could have put it in `class-filters.php` but I wanted to keep things separated and more "correct". 

Also I kinda fail to see how an interface would solve this. The point of both of these classes is to be able to reference those filter / action names in eightshift-forms without having to write them as strings each time (reusing code).

src/exception/class-unverified-request-exception.php
```diff
@@ -0,0 +1,44 @@
...
+declare( strict_types=1 );
+
+namespace Eightshift_Forms\Exception;
+
+use Eightshift_Libs\Exception\General_Exception;
+
+/**
+ * Unverified_Request_Exception class.
+ */
+class Unverified_Request_Exception extends \Exception implements General_Exception {
```
dingo-d, line 17: Usually, I like to extend a specific exception. You can look at the list here

https://www.php.net/manual/en/class.exception.php

And pick the one you think suits the use case the most (I think `RuntimeException` would be good here).

src/class-filters.php
IvanGrginovInf: This is more for documentation / keeping list of all filters / actions in a single place (and ofc so that we don't have to write them as strings each time they're used).

Personally I like when documentation lists all available hooks (filters / actions) in a single place. This is the equivalent of that in code and can later be used to automatically generate documentation for each filter / action in a single place (I'd just add documentation as constant comment and run phpdocs. Or alterntively just copy all the comments to github wiki).

Having constants added to different places using `Hooks` interface would make compiling a list of available actions / filters much more difficult (as they'd be scattered about the codebase). Think about if documentation did this? You're looking for a list of available hooks to use but they're scattered on different pages.

Also, this is the same logic by which we for example keep all custom post types in a single folder. We group things by functionality (custom post types, custom taxonomies, blocks, etc), not by use context. Hence why all filters / actions are listed in the same place, the same way as all custom post types (for example) are listed in the same place.

src/integrations/buckaroo/class-buckaroo.php
```diff
@@ -0,0 +1,571 @@
...
+    $secret_key  = \apply_filters( Filters::BUCKAROO, 'secret_key' );
+    $post        = \wp_json_encode( $post_array );
+    $md5         = md5( $post, true );
+    $post        = base64_encode( $md5 ); // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.obfuscation_base64_encode
+    $uri         = strtolower( rawurlencode( $buckaroo_uri ) );
+    $nonce       = \wp_rand( 0000000, 9999999 );
+    $time        = time();
+
+    $hmac = $website_key . 'POST' . $uri . $time . $nonce . $post;
+    $s    = hash_hmac( 'sha256', $hmac, $secret_key, true );
```
dingo-d, line 245: `$s` is a really bad variable name.

src/integrations/buckaroo/class-buckaroo.php
```diff
@@ -0,0 +1,571 @@
...
+  /**
+   * Set type of payment. Defaults to iDEAL.
+   *
+   * @param  string $pay_type  Type of payment. Defaults to iDEAL.
+   *
+   * @return  self
+   */
+  public function set_pay_type( string $pay_type ) {
+    $this->pay_type = $pay_type;
+    return $this;
```
dingo-d, line 482: So the `set_pay_type()` returns the entire class? Why?

src/class-main.php
IvanGrginovInf: I'd say implementation details, you can just call up $main->set_test(true / false) in your `BaseTest` class as needed (for example WPBrowser tests could extend `BaseWPBrowserTest` class which just runs `$main->set_test(true)` in it's `_inject()` or wherever if needed).

src/integrations/buckaroo/class-buckaroo.php
```diff
@@ -0,0 +1,571 @@
...
+   * Set return URL after payment.
+   *
+   * @param  string $return_url  Return URL after payment.
+   *
+   * @return  self
+   */
+  public function set_return_url( string $return_url ) {
+    $this->return_url = $return_url;
+
+    return $this;
```
dingo-d, line 504: Same here. Is this used for chaining purposes?

src/integrations/buckaroo/class-buckaroo.php
```diff
@@ -0,0 +1,571 @@
...
+  /**
+   * Set return URL after payment cancel.
+   *
+   * @param  string $return_url_cancel  Return URL after payment cancel.
+   * @return  self
+   */
+  public function set_return_url_cancel( string $return_url_cancel ) {
+    $this->return_url_cancel = $return_url_cancel;
+
+    return $this;
```
dingo-d, line 525: ⬆️

src/integrations/buckaroo/class-buckaroo.php
```diff
@@ -0,0 +1,571 @@
...
+   * Set return URL after payment error.
+   *
+   * @param  string $return_url_error  Return URL after payment error.
+   *
+   * @return  self
+   */
+  public function set_return_url_error( string $return_url_error ) {
+    $this->return_url_error = $return_url_error;
+
+    return $this;
```
dingo-d, line 547: :arrow_up:

src/integrations/buckaroo/class-buckaroo.php
```diff
@@ -0,0 +1,571 @@
...
+   * Set return URL after payment reject.
+   *
+   * @param  string $return_url_reject  Return URL after payment reject.
+   *
+   * @return  self
+   */
+  public function set_return_url_reject( string $return_url_reject ) {
+    $this->return_url_reject = $return_url_reject;
+
+    return $this;
```
dingo-d, line 569: :arrow_up:

src/integrations/buckaroo/class-buckaroo.php
```diff
@@ -0,0 +1,571 @@
...
+  }
```
dingo-d, line 570: `public` methods should be before `private` and `protected` ones. 

And only use `protected` if you want to extend the class, otherwise use `private`.

src/integrations/buckaroo/exceptions/class-buckaroo-request-exception.php
```diff
@@ -0,0 +1,74 @@
...
+declare( strict_types=1 );
+
+namespace Eightshift_Forms\Integrations\Buckaroo\Exceptions;
+
+use Eightshift_Libs\Exception\General_Exception;
+
+/**
+ * Buckaroo_Request_Exception class.
+ */
+class Buckaroo_Request_Exception extends \Exception implements General_Exception {
```
dingo-d, line 17: Generic exception, consider using something more specific.

src/routes/class-base-buckaroo-route.php
```diff
@@ -0,0 +1,154 @@
...
+  /**
+   * Method that returns rest response
+   *
+   * @param  \WP_REST_Request $request Data got from endpoint url.
+   *
+   * @return WP_REST_Response|mixed If response generated an error, WP_Error, if response
+   *                                is already an instance, WP_HTTP_Response, otherwise
+   *                                returns a new WP_REST_Response instance.
+   */
+  abstract public function route_callback( \WP_REST_Request $request );
```
dingo-d, line 152: Shouldn't this be defined by the REST callable interface?

src/class-main.php
IvanGrginovInf: Yup, I use a base class for tests which (if it needs injection) does this (in order to get the container for tests). For example:

```php
abstract class BaseRouteTest extends BaseTest {

  protected function _inject(Main $main, HMAC $hmac)
  {
    $main->set_test(true);
    $this->di_container = $main->build_di_container();
    ...

```

src/routes/class-base-buckaroo-route.php
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  /**
+   * Set Buckaroo to test mode if test param provided.
+   *
+   * @param array $params Request params.
+   * @return void
+   */
+  protected function set_test_if_needed( array $params ): void {
+    if ( isset( $params[ self::TEST_PARAM ] ) && filter_var( $params[ self::TEST_PARAM ], FILTER_VALIDATE_BOOL ) ) {
```
dingo-d, line 129: What's the usage of `filter_var( $params[ self::TEST_PARAM ], FILTER_VALIDATE_BOOL )` for?

src/integrations/buckaroo/class-buckaroo.php
IvanGrginovInf: Good point, I copied the Buckaroo's authorization example and forgot to change this after.

src/routes/class-base-route.php
```diff
@@ -79,80 +94,288 @@ protected function get_route_name() : string {
...
-   * @return mixed
+   * @param  \WP_REST_Request $request WP_REST_Request object.
+   * @param  string           $required_filter (Optional) Filter that needs to exist to verify this request.
+   *
+   * @throws Unverified_Request_Exception When we should abort the request for some reason.
+   *
+   * @return array            filtered request params.
    */
-  protected function rest_response_handler_unknown_error( array $data = array() ) {
+  protected function verify_request( \WP_REST_Request $request, string $required_filter = '' ): array {
```
dingo-d, line 114: Why is this in the base abstract class? It would make more sense if it's in a separate class IMO. Because this way, all the classes extending the base route class will have this method available to it. And they may not use it, which kinda smells like a SRP violation to me.

src/routes/class-base-route.php
```diff
@@ -79,80 +94,288 @@ protected function get_route_name() : string {
...
    * Checks if all required parameters are present in request.
    *
    * @param  array $parameters Array of request parameters.
+   * @param  bool  $is_post    (Optional) True if we're checking POST params instead of GET params.
    * @return array Returns array of missing parameters to pass in response.
    */
-  protected function find_required_missing_params( array $parameters ): array {
-    $missing_params = [];
-    foreach ( $this->get_required_missing_params() as $required_param ) {
+  protected function find_required_missing_params( array $parameters, bool $is_post = false ): array {
```
dingo-d, line 170: This also looks kinda like it should be a `private` no? Or is it used in the extended classes?

src/integrations/buckaroo/class-buckaroo.php
IvanGrginovInf: Yeah not sure, I use an [VSCode extension for generating setters / getters](https://github.com/phproberto/vscode-php-getters-setters) and this is how it generates them, just didn't bother changing it. 🤷‍♂️  I would imagine this is so that you can chain setter calls like so:

```php
$class->set_something('something')->set_something_else('something else')...;
```

I can remove this if you like, it's not really used.

src/integrations/buckaroo/class-buckaroo.php
IvanGrginovInf: See comment above ⬆️

src/routes/class-base-route.php
```diff
@@ -79,80 +94,288 @@ protected function get_route_name() : string {
...
+      'buckaroo-missing-keys' => [
+        'code' => 400,
+        'message' => esc_html__( 'Not all Buckaroo keys are set', 'eightshift-forms' ),
+      ],
+      'buckaroo-request-exception' => [
+        'code' => 400,
+        'message' => esc_html__( 'Error ocurred, unable to redirect to Buckaroo', 'eightshift-forms' ),
+      ],
+    ];
+  }
```
dingo-d, line 373: This entire class seems a bit overloaded for a template.

The idea of base classes is to be the template on which **all other** classes can be build from and reduce duplication. To me, it looks like this base class does waaaay too much. That is, I could see a rest class that wouldn't use any of these, and that kinda defeats the purpose of the template method pattern.

src/integrations/buckaroo/class-buckaroo.php
IvanGrginovInf: Yeah I was thinking about this. But does it make sense to have setters / getters on top? I always put them on bottom because, you know, you don't really need to look at them 99% of the time. I dunno, I also agree that we should have order (public methods first, then protected then private), not really sure what is better here.

src/routes/class-test-route.php
```diff
@@ -0,0 +1,129 @@
+<?php
+/**
```
dingo-d, line 2: Testing as in, healthcheck test or used in automated testing?

Because any mock or a fixture should be contained in the `tests` folder. This is exposed to public use (without it doing anything really).

src/routes/class-base-buckaroo-route.php
IvanGrginovInf: Yes, this is the abstract route for routes that handle Buckaroo things which extends `Base_Route` (which implements `Callable_Route`). Because of this it needs to have a `route_callback` defined (but as `abstract` because every Buckaroo route that extends this one will have a different `route_callback` implementation).

tests/Integrations/Buckaroo/DataProvider.php
```diff
@@ -0,0 +1,136 @@
...
+      'BRQ_SERVICE_IDEAL_CONSUMERISSUER' => 'ABN AMRO',
+      'BRQ_SERVICE_IDEAL_TRANSACTIONID' => '0000000000000001',
+      'BRQ_STATUSCODE' => 791,
+      'BRQ_STATUSMESSAGE' => 'Pending processing',
+      'BRQ_TEST' => true,
+      'BRQ_TIMESTAMP' => '2020-10-08 11:01:13',
+      'BRQ_TRANSACTIONS' => 'AAC90995D35347C9B2D8D9EDFB265B1F',
+      'BRQ_WEBSITEKEY' => 'pjTB7CdkZ8',
+      'BRQ_SIGNATURE' => '0a03229db3842c7042fbca3091092d8df628f097',
+    ];
```
dingo-d, line 134: Liking this very much! 👍

dingo-d (changes requested):
I've left some comments and questions. Would love to hear the reasoning behind some 👍

src/routes/class-base-buckaroo-route.php
IvanGrginovInf: This is used to allow triggering a test Buckaroo response from Form editor (by adding a hidden field `test` with value `1`). Rather than requiring that the value is `=== 1`, by using this filter the editor can set it however they like and it will work. It will work if they put the value as `1`, ` 1`, `yes`, `true`, etc.

src/routes/class-base-route.php
IvanGrginovInf: This is actually seriously useful and cuts down on repeated code A LOT. All routes should (and they do) use it as the first thing that happens in their `route_callback`. This does a number of things that were previously repeated in the code.

This method does things like:
- If the route requires a filter to do it's thing (which needs to be defined in the project) - it verifies that it exists (`has_filter()`)
- If the route has required parameters - it will check that they are all set
- If route requires authorization - it will check that the authorization parameter is added and correct for this request
- It will check captcha is correct (if used by route)

And you can easily set which of these things are used depending on some easy functions you can set per route (for example you have `get_required_params(): array` method. You can return 1 or more parameter names here and if you do, `verify_request()` will check if they're set and if not throw an exception with the proper response set (which we then catch in `route_callback` and return the response)

I'm actually really proud of this functionality, it allows you to cut back dramatically on boilerplate in routes and to dictate how the route behaves and to enable complex functionality for such route (like requiring authorization) with easy to understand methods / parameters.

src/routes/class-base-route.php
IvanGrginovInf: Yup good point, can be private.

src/blocks/custom/form/form.php
dingo-d: Oh, I see now 👍  Then it's ok 🙂

src/class-actions.php
dingo-d: To me, these look kinda like enums, even though their use case is a bit different. 

I just see classes as encapsulated code that holds logic. Constants are just names without any logic whatsoever. In that aspect, an interface makes more sense. No logic, just a contract and you have encapsulation. You can implement numerous different interfaces in different places where it makes sense. Like `RouteInterface` in the libs. It only holds the constants for the HTTP verbs.

Also, these constants should all have visibility modifiers next to them (missed that in the first round) 😄

src/class-filters.php
dingo-d: You can look at my response above. Keeping them in one place is fine. I just think interface or enum makes more sense for this instead of a class.

src/integrations/buckaroo/class-buckaroo.php
dingo-d: I mean, I don't think it adds some super extra overhead so it can stay. I'm just not used to see it in WP I guess (I've seen this in Laravel tho).

src/routes/class-base-route.php
IvanGrginovInf: Dunno, all routes in forms use almost all functionality here, like before all this code was in each separate class and there was a lot of code duplication.

src/routes/class-test-route.php
IvanGrginovInf: Routes for automated tests, it's only present in di container used for tests. Not sure how else I'd add a route like this. I can put it in `tests` folder and include it in `class-main.php` for tests.

src/integrations/buckaroo/class-buckaroo.php
dingo-d: it adds an unnecessary boilerplate, but it's a common practice to have them on top. Plus `public`, `protected`, `private` is also common in most places I've seen (I know at WPCS it's mandatory).

I've seen a package in one of the PHP meetups that kinda extracts getters and setters, cannot remember what was the name :/ But it reduces boilerplate code a lot

src/routes/class-base-buckaroo-route.php
dingo-d: Cool!

tests/Integrations/Buckaroo/DataProvider.php
IvanGrginovInf: Currently not really used but when I'll be writing tests for Buckaroo routes specifically it will definitely be used.

src/routes/class-base-route.php
dingo-d: I guess I'm not used to seeing it being built like that. Regular WP rest has a security callback that gets called before the request is executed (but should be only used for authorization checks), and they usually verify things using schema.

https://developer.wordpress.org/rest-api/extending-the-rest-api/schema/#resource-schema

I won't argue about its usefulness, just looked oddly specific for a general base rout 🤷‍♂️

IvanGrginovInf: Yeah I agree that schema would be useful, didn't know that existed TBH (tho not sure if you specifiy different schema for POST vs GET, which is something we need in `buckaroo-response-handler` route - as that routes has (different) required GET and POST parameters.

At this point I'm on a really tight deadline (and we have to add Mailchimp as well as a new feature request while the deadline stays the same) so I'd rather not refactor this for now.

src/integrations/buckaroo/class-buckaroo.php
IvanGrginovInf: Ok, sounds good, will move them to top!

src/routes/class-base-route.php
dingo-d: No, it's ok, leave it in. You have tests so you can always refactor later :trollface:

src/routes/class-test-route.php
dingo-d: You can leave it in for now like this (if it's not registered in the regular flow it's ok). 

It's just that usually in `src` you'll put anything that you'll use in the production, not tests.

dingo-d (approved):

iruzevic (approved):

PR #14, opened 2020-10-22
Multi type forms

IvanGrginovInf:
Changelog
---
* Added ability for form to do multiple things
* Added a new Form type - Custom Event(s) - it can fire custom JS events (set in editor)
* Rework of how form theme works (it's now selected on form render view, rather than on form itself - this way we can have the same form with two difference themes in 2 different places)


src/blocks/components/components-panel/components-panel-editor.scss
```diff
@@ -0,0 +1,8 @@
+.components-panel {
+  .components-notice {
+    &.is-dismissible {
+      padding-right: 0;
+      margin: 10px 0;
+    }
+  }
+}
```
dingo-d, line 8: Should the notice be its own element? Like in a standalone file?

A panel is a container. Or did you meant

```suggestion
.components-panel {
  &__notice {
    &.is-dismissible {
      padding-right: 0;
      margin: 10px 0;
    }
  }
}
```

src/blocks/components/form-error-message-wrapper/form-error-message-wrapper-style.scss
```diff
@@ -0,0 +1,5 @@
+.form-error-message-wrapper {
+  &.hide-form-message {
```
dingo-d, line 2: Why not `.is-hidden`?

src/blocks/components/form-error-message-wrapper/form-error-message-wrapper.php
```diff
@@ -0,0 +1,23 @@
...
+
+use Eightshift_Libs\Helpers\Components;
+
+$block_class     = $attributes['blockClass'] ?? '';
+$component_class = 'form-error-message-wrapper';
+
+$block_classes = Components::classnames([
+  $component_class,
+  "js-{$component_class}",
+  'hide-form-message',
```
dingo-d, line 18: This is related to the comment above 😄

src/blocks/custom/form/assets/form.js
```diff
@@ -37,59 +54,158 @@ export class Form {
...
-    this.CLASS_HIDE_MESSAGE = 'hide-form-message';
-    this.CLASS_HIDE_OVERLAY = 'hide-form-overlay';
+    this.errorMessageClasses = [
+      'form-message',
+      'js-form-message',
+      'js-form-message--error',
+      'form-message__type--error',
+    ];
+
+    this.errors = [];
```
dingo-d, line 64: download

That's a big ass constructor 😬

You sure this class is not doing too much? 😅 

I mean if it works...

src/blocks/components/components-panel/components-panel-editor.scss
IvanGrginovInf: Notice is default Gutenberg component (https://developer.wordpress.org/block-editor/components/notice/) and it looks ugly when used in sidebar.

But it does make sense to add this style as the style of that component (`components/notice/notice-style-editor.scss`), will do that! 👍

src/blocks/components/form-error-message-wrapper/form-error-message-wrapper-style.scss
IvanGrginovInf: Because previously when discussing as similar PR with @iruzevic he metioned we should probably use specific classes for this for components (rather than something general as `is-hidden`). I'm kinda fine with both 🤷‍♂️

src/blocks/components/form-error-message-wrapper/form-error-message-wrapper.php
IvanGrginovInf: See comment above

src/blocks/custom/form/assets/form.js
IvanGrginovInf: Yeah it handles all logic for submitting forms, it is long I agree, and maybe I could refactor it to break it up into little pieces but I'd rather not do that right now 😅

src/blocks/custom/forms/forms.php
```diff
@@ -8,12 +8,32 @@
...
+
+if ( ! empty( $theme ) ) {
+  $post_blocks = Forms::recursively_change_theme_for_all_blocks( parse_blocks( $post_content ), $theme );
+} else {
+  $post_blocks = parse_blocks( $post_content );
+}
+
+echo '<pre>';
+echo varexport( $post_blocks );
+echo '</pre>';
```
dingo-d, line 36: What's this for?

src/helpers/class-forms.php
```diff
@@ -28,4 +28,46 @@ public static function maybe_override_value_from_query_string( string $value, st
...
+   * @param  string $form_type                   Used form type (used if not complex).
+   * @param  array  $form_types_complex          Used form types.
+   * @param  array  $form_types_complex_redirect Used form types that redirect on success.
+   * @return array
+   */
+  public static function detect_used_types( bool $is_complex, string $form_type, array $form_types_complex, array $form_types_complex_redirect ): array {
+    $used_types = [];
+
+    if ( $is_complex ) {
+      foreach ( array_merge( $form_types_complex, $form_types_complex_redirect ) as $complex_form_type ) {
```
dingo-d, line 45: Eww, this is really not readable 😣

Can you merge arrays outside and then foreach them. The memory impact of assigning an extra variable should be negligible.

src/blocks/custom/forms/forms.php
IvanGrginovInf: Stray deubgging stuff, removed!

dingo-d (changes requested):
Left some comments. Looks good otherwise (some parts are a bit like the Chinese alphabet to me, but I guess the logic is sound 😂 )

src/blocks/components/components-panel/components-panel-editor.scss
dingo-d: Oh, didn't know that it's a core element. They have ugly css names 😂

IvanGrginovInf:
Ty, some good stuff, fixed all the things!

src/blocks/components/form-error-message-wrapper/form-error-message-wrapper-style.scss
dingo-d: Dunno, I thought that we always used the same 'helper' css classes for these kinds of things 🤷‍♂️ 

I'm fine with both, just thought the default was like `.is-{action}` 🤷‍♂️

src/blocks/custom/form/assets/form.js
dingo-d: Yeah, if it's not broken don't fix it 😅 

But you have tests so it shouldn't be hard to refactor later on 💪

src/blocks/components/form-error-message-wrapper/form-error-message-wrapper-style.scss
IvanGrginovInf: We should discuss this on the next Protein / standup, I agree we should be consistent. Also makes sense for `is-{action}` format, I'll fix this.

dingo-d (approved):

src/blocks/components/components-panel/components-panel-editor.scss
IvanGrginovInf: Yeah, fake BEM xD

PR #27, opened 2020-11-06
Last changes before 0.1.0

IvanGrginovInf:
Changelog
---
* Better handling of One Time / Recurring Buckaroo options
* Allow setting of Buckaroo iDEAL description in form settings
* Email - Ability to send custom email to user on form submit
* Added feature to set redirect URL on success (for all form types)
* Fixed some Buckaroo redirection issues (caused by not urlencoding params)

After merging this I'll tag it as `0.1.0`

### DIFF
--- a/src/blocks/custom/form/assets/form.js
+++ b/src/blocks/custom/form/assets/form.js
@@ -23,6 +23,7 @@ export class Form {
     this.DATA_ATTR_BUCKAROO_SERVICE = 'data-buckaroo-service';
     this.DATA_ATTR_SUCCESSFULLY_SUBMITTED = 'data-form-successfully-submitted';
     this.DATA_ATTR_FIELD_DONT_SEND = 'data-do-not-send';
+    this.DATA_ATTR_REDIRECT_URL_SUCCESS = 'data-redirect-on-success';
     this.EVENT_SUBMIT = 'ef-submit';
     this.STATE_IS_LOADING = false;
     this.CLASS_FORM_SUBMITTING = 'form-submitting';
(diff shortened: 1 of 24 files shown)


iruzevic (approved):

src/blocks/custom/form/assets/form.js
```diff
@@ -115,10 +124,25 @@ export class Form {
...
       }
     });
   }
 
+  /**
+   * Redirects user on success if needed.
+   *
+   * @param {boolean} success Is form successfully submitted
+   */
+  maybeRedirect(success) {
```
gabriel-glo, line 138: This method name doesn't look good me. Why do you need to add `maybe`. It will redirect only if all conditions are met.

gabriel-glo (changes requested):

src/blocks/custom/form/assets/form.js
IvanGrginovInf: @gabriel-glo I mean that's why it's prefixed with `maybe`. If it was called just `redirect`, it runs and it doesn't redirect the user (because some conditions aren't met) then it shouldn't really be called `redirect`. 😄

kmaric69: why not call it then `onSuccessRedirect`?

dingo-d: Because it handles more than success. It handles success, if the redirect should happen or if the redirect URL for success is set 😉 

I'm fine with the name.

IvanGrginovInf: > why not call it then onSuccessRedirect?

It redirects only if:
- It's a success
- "Redirect on success" option is enabled for this particular form
- Redirect URL is set

So `onSuccessRedirect` would also be incorrect for what it does.

> Because it handles more than success. It handles success, if the redirect should happen or if the redirect URL for success is set 😉

Exactly.

IvanGrginovInf: I mean I'm open to alternative names if that name actually describes what this function does.

gabriel-glo: Only thing that comes to mind is `conditionalSuccessRedirect` but I don't think it's a better name than the current one.

gabriel-glo (approved):

dingo-d (approved):

dingo-d (approved):

dingo-d (approved):

PR #28, opened 2020-11-17
Code required for pay by email option

IvanGrginovInf:


src/routes/class-buckaroo-response-handler-route.php
```diff
@@ -142,8 +142,12 @@ public function route_callback( \WP_REST_Request $request ) {
...
 
     try {
-      if ( has_action( self::BUCKAROO_RESPONSE_HANDLER ) ) {
-        do_action( self::BUCKAROO_RESPONSE_HANDLER, $params, $buckaroo_params );
+      if ( has_filter( Filters::BUCKAROO_FILTER_BUCKAROO_PARAMS ) ) {
+        $buckaroo_params = apply_filters( Filters::BUCKAROO_FILTER_BUCKAROO_PARAMS, $params, $buckaroo_params );
+      }
+
+      if ( has_action( Actions::BUCKAROO_RESPONSE_HANDLER ) ) {
+        do_action( Actions::BUCKAROO_RESPONSE_HANDLER, $params, $buckaroo_params );
```
dingo-d, line 150: Do you need to check for filters and actions before executing them? I think you can just leave them around, no? Unless you want to ensure that you only have one filter/action applied 🤔

dingo-d (approved):
Left one comment, otherwise looks ok to me

dingo-d (approved):

src/routes/class-buckaroo-response-handler-route.php
IvanGrginovInf: Hmm not sure, will have to check. Was just going by the same logic as for filters. Tho now you mention this it makes sense that you can just run it even tho it has no callbacks, will check!

dingo-d (approved):

PR #40, opened 2021-01-19
Better Mailchimp logic

IvanGrginovInf:
Mailchimp route only adds users by default (doesn't update them).

We should never really have a form which updates on frontend (unless it's specifically a form for updating user data - which is currently out of scope).

If user already exists it still returns 200 with an appropriate message "User already exists" (as that's not really an error).

### DIFF
--- a/src/integrations/mailchimp/class-mailchimp.php
+++ b/src/integrations/mailchimp/class-mailchimp.php
@@ -69,6 +69,34 @@ class Mailchimp {
   }
 
   /**
+   * Adds a member in Mailchimp.
+   *
+   * @param  string $list_id      Audience list ID.
+   * @param  string $email        Contact's email.
+   * @param  array  $merge_fields List of merge fields to add to request.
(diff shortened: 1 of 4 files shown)


dingo-d (commented):
The error that PHPStan warns you about is something you should address. The getResponse can be null: https://github.com/guzzle/guzzle/blob/34bc199aef77ff1c13f7126c0464f8fa9ee55f9f/src/Exception/RequestException.php#L141

IvanGrginovInf:
> The error that PHPStan warns you about is something you should address
@dingo-d Yup, didn't have time to run the tests / analyze when I opened this but had to open a PR so I don't forget to merge the changes (since I was changing this in a project).

Fixed phpstan errors, fixed tests, in the process found and fixed a bug, the whole 9 yards. 😄

iruzevic (approved):

kmaric69 (approved):

dingo-d (approved):

PR #42, opened 2021-01-21
Add nonce

IvanGrginovInf:
Changelog
---
* Added nonce field to all forms
* Added base logic for all routes to verify nonce (if toggled on by the route)

### DIFF
--- a/src/routes/class-base-route.php
+++ b/src/routes/class-base-route.php
@@ -167,6 +167,19 @@ abstract class Base_Route extends Libs_Base_Route implements Callable_Route, Act
       }
     }
 
+    // Verify nonce if submitted.
+    if ( $this->does_require_nonce_verification() ) {
+      if (
+        ! isset( $params['nonce'] ) ||
+        ! isset( $params['form-unique-id'] ) ||
+        ! wp_verify_nonce( $params['nonce'], $params['form-unique-id'] )
+      ) {
+        throw new Unverified_Request_Exception(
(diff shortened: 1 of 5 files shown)


gabriel-glo (dismissed):

src/routes/class-base-route.php
```diff
@@ -284,6 +297,15 @@ protected function get_irrelevant_params(): array {
     return [];
   }
 
+  /**
+   * Toggle if this route requires nonce verification.
+   *
+   * @return bool
+   */
+  protected function does_require_nonce_verification(): bool {
```
dingo-d, line 305: The naming of this method is a bit ambiguous as to what it returns. 

https://github.com/kettanaito/naming-cheatsheet#reflect-the-expected-result

dingo-d (changes requested):
Left one comment regarding the naming of the method.

src/routes/class-base-route.php
IvanGrginovInf: Was following our naming standard for booleans (should be prefix by a question, usually `is_something` but `is` doesn't fit in here so I used `does`).

Anyway, makes sense, I'll just name it `requires_nonce_verification` then (so it ready nicely: `$this->requires_nonce_verification()`) 👍

dingo-d (approved):

kmaric69 (approved):

gabriel-glo (approved):

PR #43, opened 2021-01-25
All routes used on frontend (in forms) are now POST by default

IvanGrginovInf:
Changelog
---
* Ajax form submission uses POST
* Email route now only accepts POST requests + added nonce verification
* Mailchimp route now only accepts POST requests + added nonce verification
* CRM (add entity) route now only accepts POST requests + added nonce verification
* Buckaroo route now only accepts POST requests + added nonce verification

### DIFF
--- a/src/routes/class-base-route.php
+++ b/src/routes/class-base-route.php
@@ -123,10 +123,19 @@ abstract class Base_Route extends Libs_Base_Route implements Callable_Route, Act
    */
   protected function get_callback_arguments() : array {
     return [
-      'methods'  => static::READABLE,
+      'methods'  => $this->get_methods(),
       'callback' => [ $this, 'route_callback' ],
       'permission_callback' => [ $this, 'permission_callback' ],
     ];
+  }
+
+  /**
+   * Returns allowed methods for this route.
+   *
+   * @return string|array
+   */
+  protected function get_methods() {
+    return static::READABLE;
   }
 
   /**
--- a/src/routes/class-mailchimp-route.php
+++ b/src/routes/class-mailchimp-route.php
@@ -200,4 +200,13 @@ class Mailchimp_Route extends Base_Route implements Filters {
   protected function requires_nonce_verification(): bool {
     return true;
   }
+
+  /**
+   * Returns allowed methods for this route.
+   *
+   * @return mixed
+   */
+  protected function get_methods() {
+    return static::CREATABLE;
+  }
 }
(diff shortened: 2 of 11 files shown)


kmaric69 (approved):

src/routes/class-base-route.php
```diff
@@ -123,12 +123,21 @@ protected function get_route_name() : string {
...
       'permission_callback' => [ $this, 'permission_callback' ],
     ];
   }
 
+  /**
+   * Returns allowed methods for this route.
+   *
+   * @return string|array
+   */
+  protected function get_methods() {
```
dingo-d, line 137: Why method? I'd leave this as a constant, that's the way we always do it. If you need multiple methods, you can set them as a constant either as an array, or a comma-separated string.

src/routes/class-mailchimp-route.php
```diff
@@ -200,4 +200,13 @@ protected function get_irrelevant_params(): array {
   protected function requires_nonce_verification(): bool {
     return true;
   }
+
+  /**
+   * Returns allowed methods for this route.
+   *
+   * @return mixed
```
dingo-d, line 207: Fix the return type here to be the same as in the other methods.

tests/Routes/MailchimpRouteTest.php
```diff
@@ -25,8 +25,8 @@ protected function _before()
    */
   public function testRestCallSuccessfulWhenAddingNewMembers()
   {
-    $request = new \WP_REST_Request('GET', $this->route_endpoint->get_route_uri());
-    $request->params['GET'] = [
+    $request = new \WP_REST_Request('POST', $this->route_endpoint->get_route_uri());
```
dingo-d, line 28: You could extract this `'POST'` string in a constant and then just changed it in one place, the constant would remain the same in other places but referring to the `'POST'` string.

dingo-d (changes requested):
Left some comments and suggestions

tests/Routes/MailchimpRouteTest.php
IvanGrginovInf: I need to be able to mock both GET and POST requests here (this is a mock WP_REST_Request). Tho I do agree since it's the same for all tests in this file, it could just be extracted into a const. 👍

gabriel-glo (dismissed):

iruzevic (dismissed):

dingo-d (approved):

dingo-d (approved):

gabriel-glo (approved):

PR #44, opened 2021-01-27
Sanitize all fields in all routes by default

IvanGrginovInf:


iruzevic (dismissed):

src/routes/class-base-route.php
```diff
@@ -490,4 +490,24 @@ private function fix_dot_underscore_replacement( array $params ): array {
...
   }
+
+  /**
+   * Sanitizes all received fields recursively. If a field is something we don't need to
+   * sanitize then we don't touch it.
+   *
+   * @param  array $params Array of params.
+   * @return array
+   */
+  private function sanitize_fields( array $params ) {
```
dingo-d, line 501: This looks like it could be relatively easy to unit test. I'm a bit nervous seeing recursions inside a loop 😅

IvanGrginovInf: Would need to make it public then (no problem with that tho you know, you don't really want anyone modifying this too much).

Also since in tests we use stubs for `wp_unslash()` and `sanitize_text_fields()` I don't think a unit test would do much good except prove that recursion doesn't end up causing an infinite loop (tho I'd be happy to write it, just out of all methods, this is the last one I'd want to make `public` :D)

dingo-d: To test this, you can mock the implementation that will extend the base route class, and create a dummy method that will use this method (plus mock the WP functions). Or write integration tests for that. 

The only concern is the infinite loop and that all edge cases are covered (pass different types in the `$params` array).

IvanGrginovInf: Makes sense, will do!

IvanGrginovInf: > mock the implementation that will extend the base route class, and create a dummy method that will use this method

@dingo-d  Tho how will I do that if `sanitize_fields()` is `private`? 😅  It would at least need to be `protected` then, no?

dingo-d: Well, technically, you need to test the `verify_request` method and make sure that params are properly escaped, no matter whatever they contain. That should be mockable I think.

IvanGrginovInf: Oh, yeah that makes sense 🤦 got it, will do!

kmaric69 (dismissed):

gabriel-glo (dismissed):

iruzevic (approved):

dingo-d (approved):

PR #51, opened 2021-02-11
Feature/mailerlite integration

iruzevic:
- added mailerlite integration
- WIP: Tests


src/blocks/custom/form/components/form-mailerlite-options.js
```diff
@@ -0,0 +1,41 @@
...
+    },
+    ...groups,
+  ];
+
+  return (
+    <Fragment>
+      {onChangeGroupId &&
+        <SelectControl
+          label={__('Group ID', 'eightshift-forms')}
+          help={__('Please select which group does this form add members to', 'eightshift-forms')}
```
dingo-d, line 31:
```suggestion
          help={__('Please select to which group does this form add members to', 'eightshift-forms')}
```

src/blocks/custom/form/components/form-options.js
```diff
@@ -326,6 +346,22 @@ export const FormOptions = (props) => {
 
               </Fragment>
             )}
+            {tab.name === 'mailerlite' && (
+              <Fragment>
+                <br />
+                <strong className="notice-title">{__('MailerLite Options', 'eightshift-forms')}</strong>
+                <p>{__('These are options for when your form is sending data to MailerLite.', 'eightshift-forms')}</p>
```
dingo-d, line 353:
```suggestion
                <p>{__('These are the options for when your form is sending data to MailerLite.', 'eightshift-forms')}</p>
```

src/enqueue/class-localization-constants.php
```diff
@@ -101,6 +118,8 @@ public function __construct(
     $this->send_email_route            = $send_email_route;
     $this->mailchimp_route             = $mailchimp_route;
     $this->mailchimp                   = $mailchimp;
+    $this->mailerlite_route             = $mailerlite_route;
+    $this->mailerlite                   = $mailerlite;
```
dingo-d, line 122:
```suggestion
    $this->mailerlite_route            = $mailerlite_route;
    $this->mailerlite                  = $mailerlite;
```

I think forms are using older CS where we used to align all the equal signs (like WP does). These seem to be offset by one space to the right.

src/enqueue/class-localization-constants.php
```diff
@@ -219,6 +243,20 @@ protected function add_mailchimp_constants( array $localization ): array {
     return $localization;
   }
 
+  /**
+   * Localize all constants required for Mailerlite integration.
+   *
+   * @param  array $localization Existing localizations.
+   * @return array
+   */
+  protected function add_mailerlite_constants( array $localization ): array {
```
dingo-d, line 252: If the class isn't used for extending, all of these should be `private` not `protected`.

https://medium.com/@ivorobioff/keep-it-private-unless-you-guarantee-to-protect-it-1b95382c1961

src/hooks/class-filters.php
```diff
@@ -216,6 +216,29 @@ interface Filters {
...
   const MAILCHIMP = 'eightshift_forms/mailchimp';
 
+  /**
+   * Filter used for providing mailerlite credentials.
+   * IMPORTANT - Make sure to always return a string, even if $key isn't set.
+   *
+   * Example:
+   *
+   *   public function register(): void {
+   *     add_filter( 'eightshift_forms/mailerlite', [ $this, 'get_info' ], 1, 1 );
```
dingo-d, line 226:
```suggestion
   *     add_filter( 'eightshift_forms/mailerlite', [ $this, 'get_info' ], 1 );
```

The 1 can be omitted when only one argument is passed to the filter.

src/integrations/mailerlite/class-mailerlite-client.php
```diff
@@ -0,0 +1,51 @@
...
+   *
+   * @return void
+   */
+  public function set_config() {
+    $api_key = \has_filter( Filters::MAILERLITE ) ? \apply_filters( Filters::MAILERLITE, 'api_key' ) : '';
+
+    $guzzle = new GuzzleHttp();
+    $guzzleClient = new Guzzle6($guzzle);
+
+    $this->client = new MailerLiteClient($api_key, $guzzleClient);
```
dingo-d, line 40: Why is there a direct call towards these clients here? Shouldn't this be abstracted out and added using DI?

src/integrations/mailerlite/class-mailerlite-client.php
```diff
@@ -0,0 +1,51 @@
...
+ * Mailerlite integration class.
+ */
+class Mailerlite_Client implements Mailerlite_Client_Interface {
+
+  /**
+   * Mailerlite client object.
+   *
+   * @var MailerLiteClient
+   */
+  protected $client;
```
dingo-d, line 27: Why `protected`?

src/integrations/mailerlite/class-mailerlite.php
```diff
@@ -0,0 +1,98 @@
...
+   * @var Mailerlite_Client_Interface
+   */
+  private $mailerlite_client;
+
+  /**
+   * Constructs object
+   *
+   * @param Mailerlite_Client_Interface $mailerlite_client Mailerlite client.
+   */
+  public function __construct( Mailerlite_Client_Interface $mailerlite_client ) {
```
dingo-d, line 39: I reckon that all these interfaces look the same (just getter and setter).

Interfaces should be abstract on purpose. So it makes no sense to have two same interfaces but with specific names (`Mailerlite_Client_Interface` and `Mailchimp_Marketing_Client_Interface`). Defeats the purpose of using the interface.

It would be better to just have one `ClientInterface` which has the methods that can be implemented in a different way by `Mailerlite` and `Mailchimp` client classes. Then when injecting to classes, the type hint is against a general interface, but the actual implementation will be the classes implementing that interface.

src/integrations/mailerlite/class-mailerlite.php
```diff
@@ -0,0 +1,98 @@
...
+   * @param  string $email           Contact's email.
+   * @param  array  $subscriber_data List of merge fields to add to request.
+   * @param  array  $params          Additional params
+   * @return mixed
+   *
+   * @throws \Exception When response is invalid.
+   */
+  public function add_subscriber( string $group_id, string $email, array $subscriber_data, array $params = [] ) {
+    $this->setup_client_config_and_verify();
+    error_log( print_r( ( $subscriber_data ), true ) );
```
dingo-d, line 67: error log left here. Not sure why PHPCS didn't pick this up.

dingo-d (changes requested):
Left a few comments.

src/integrations/mailerlite/class-mailerlite-client.php
iruzevic: This is because of this specific implementation only on this integration.
https://github.com/mailerlite/mailerlite-api-v2-php-sdk#use-your-preferred-http-client

I didn't see the point of extracting this into abstraction

src/integrations/mailerlite/class-mailerlite-client.php
iruzevic: Not sure, I just copied it from the mailchimp integration. Should it be private?

src/integrations/mailerlite/class-mailerlite.php
iruzevic: Agreed

gabriel-glo (approved):

src/blocks/class-blocks.php
```diff
@@ -37,10 +37,20 @@ public function register() {
    */
   public function get_all_allowed_forms_blocks( $allowed_block_types, $post ) {
     if ( $post->post_type === Forms::POST_TYPE_SLUG ) {
+
+      // Remove forms select on form builder post type.
+      $allInternalBlock = $this->get_all_blocks_list();
+      if (($key = array_search('eightshift-forms/forms', $allInternalBlock, true)) !== false) {
```
dingo-d, line 43: Assignment inside a conditional should be avoided at all costs. It's hard to read and prone to errors.

The assignment to `$key` should be extracted out of the conditional.

Also, I'm not 100% sure what the `$allInternalBlock` returns, but if it's a simple array,
```php
isset($allInternalBlock['eightshift-forms/forms'])
``` 
should do the trick. Plus it's way faster than an `array_search()` call.

dingo-d (changes requested):
Left one comment.

IvanGrginovInf (changes requested):

IvanGrginovInf:
Fixed all PR comments and added tests.

IvanGrginovInf (approved):

PR #59, opened 2021-05-25
Install node-sass to fix GH actions error

IvanGrginovInf:


package.json
```diff
@@ -33,6 +33,7 @@
   },
   "devDependencies": {
     "husky": "^4.3.6",
+    "node-sass": "^5.0.0",
```
iruzevic, line 36: could you go with the dart-sass that we have in flibs?

IvanGrginovInf: Build is failing with error, trying to fix it, it doesn't happen for me locally: https://github.com/infinum/eightshift-forms/runs/2663211028?check_suite_focus=true#step:9:8. Tho now that I think about it, there might be another way to fix it.

IvanGrginovInf: Just updated all packages + nuked package.lock, should be ok now (since `sass` is a dependency of `frontend-libs`).

package.json
```diff
@@ -34,11 +34,11 @@
...
     "husky": "^4.3.6",
     "webpack": "^4.42.0",
-    "webpack-cli": "^4.3.1"
+    "webpack-cli": "^4.7.0"
   },
   "dependencies": {
-    "@eightshift/frontend-libs": "https://github.com/infinum/eightshift-frontend-libs#develop",
-    "@wordpress/server-side-render": "^1.16.0"
+    "@eightshift/frontend-libs": "git+https://github.com/infinum/eightshift-frontend-libs.git#develop",
+    "@wordpress/server-side-render": "^1.21.3"
```
iruzevic, line 41: this should work without this lib because it is instaled in the core. can you please test

iruzevic (changes requested):

package.json
IvanGrginovInf: Will test in another PR, just trying to get release zipping working now on releases.

iruzevic (approved):

goranalkovic-infinum (approved):

PR #64, opened 2021-07-27
Adding Greenhouse Integration

iruzevic:
## Added
- GH post application
- GH get job/jobs
- Uploads helper
- Validation helper
- Test coverage

## Tests
Screenshot 2021-07-27 at 11 59 50
Screenshot 2021-07-27 at 12 20 44


iruzevic:
Fixed phpstan and linting issues

IvanGrginovInf (dismissed):

src/Helpers/Validation.php
```diff
@@ -0,0 +1,84 @@
...
+	/**
+	 * Validate File Minimum size.
+	 *
+	 * @param integer $fileSize File size value.
+	 *
+	 * @return boolean
+	 */
+	public static function isFileMaxSizeValid(int $fileSize): bool
+	{
+		return $fileSize <= static::MAX_FILE_SIZE ? true : false;
```
mbmjertan, line 54: You don't need a ternary expression here: `return $fileSize <= static::MAX_FILE_SIZE` would be enough. See https://3v4l.org/t02XW

For all comparison operators apart from the spaceship operator, the result of a comparison is [always a bool](https://3v4l.org/t02XW), so this seems just a bit redundant. Not really a (performance) issue as PHP should automatically optimize this, though.

dingo-d, line 54:
```suggestion
		return $fileSize <= static::MAX_FILE_SIZE;
```

src/Helpers/Validation.php
```diff
@@ -0,0 +1,84 @@
...
+	/**
+	 * Validate File Minimum size.
+	 *
+	 * @param integer $fileSize File size value.
+	 *
+	 * @return boolean
+	 */
+	public static function isFileMinSizeValid(int $fileSize): bool
+	{
+		return $fileSize >= static::MIN_FILE_SIZE ? true : false;
```
dingo-d, line 66:
```suggestion
		return $fileSize >= static::MIN_FILE_SIZE;
```

src/Helpers/Validation.php
```diff
@@ -0,0 +1,84 @@
...
+	 * @param string $fileTypes String of all file types.
+	 *
+	 * @return boolean
+	 */
+	public static function isFileTypeValid(string $fileName, string $fileTypes): bool
+	{
+		$fileExtension = explode('.', $fileName);
+		$validTypes = explode(',', str_replace(' ', '', $fileTypes));
+
+		return in_array(end($fileExtension), $validTypes, true) ? true : false;
```
dingo-d, line 82:
```suggestion
		return in_array(end($fileExtension), $validTypes, true);
```

src/Integrations/Greenhouse/Greenhouse.php
```diff
@@ -0,0 +1,438 @@
...
+				'body' => wp_json_encode($body),
+				'data_format' => 'body',
+			]
+		);
+
+		if (is_wp_error($response)) {
+			return [];
+		}
+
+		$body = $response['body'] ?? '';
```
dingo-d, line 118: I think you could use `wp_remote_retrieve_body($response)` here.

https://developer.wordpress.org/reference/functions/wp_remote_retrieve_body/

src/Integrations/Greenhouse/Greenhouse.php
```diff
@@ -0,0 +1,438 @@
...
+			[
+				'headers' => $this->getHeaders(),
+			]
+		);
+
+		if (is_wp_error($response)) {
+			return [];
+		}
+
+		$body = $response['body'] ?? '';
```
dingo-d, line 247: Same as the above about the remote body wp function

src/Integrations/Greenhouse/Greenhouse.php
```diff
@@ -0,0 +1,438 @@
...
```
dingo-d, line 278: ⬆️

src/Integrations/Greenhouse/Greenhouse.php
iruzevic: Fixed

src/Integrations/Greenhouse/Greenhouse.php
```diff
@@ -0,0 +1,438 @@
...
+	protected function getJobsRaw(): array
+	{
+		$output = array_map(
+			function ($job) {
+				$jobId = $job['id'];
+
+				$job = $this->getGreenhouseJob((string) $jobId);
+
+				if (!$job) {
+					return;
```
dingo-d, line 281:
```suggestion
					return [];
```

src/Rest/GreenhouseRoute.php
```diff
@@ -0,0 +1,668 @@
...
+	/**
+	 * Returns allowed methods for this route.
+	 *
+	 * @return string|array
+	 */
+	protected function getMethods()
+	{
+		return static::CREATABLE;
+	}
+}
```
dingo-d, line 668: Couldn't find anything wrong with this other than it's breaking the single responsibility principle big time. This class should handle just the REST route, everything else (sending emails, validation, etc.) should have been a different class that does just that and injected here.

I won't require any rewrite because that would take too much time, but just something to be aware of.

tests/BaseTest.php
```diff
@@ -76,6 +77,9 @@ protected function _before()
 				'wp_verify_nonce' => function () {
 					return true;
 				},
+				'is_email' => function (string $email) {
+					return $email === 'invalid' ? false : true;
```
dingo-d, line 81:
```suggestion
					return $email === 'invalid';
```

tests/Integrations/Greenhouse/GreenhouseTest.php
```diff
@@ -0,0 +1,501 @@
...
+	protected function _inject(DataProvider $dataProvider)
+	{
+		$this->dataProvider = $dataProvider;
+	}
+
+	protected function _before()
+	{
+		parent::_before();
+		$this->greenhouse = $this->diContainer->get(MockGreenhouse::class);
+	}
```
dingo-d, line 24: I think you should add `_after()` to cleanup your test.

tests/Integrations/Greenhouse/GreenhouseTest.php
```diff
@@ -0,0 +1,501 @@
...
+		);
+
+		$params = DataProvider::greenhouseApplicationParams();
+
+		$response = $this->greenhouse->postGreenhouseApplication(
+			DataProvider::MOCKED_GH_APPLICATION_SUCCESS_URL,
+			$params
+		);
+
+		$this->assertEquals($response['success'], 'Candidate saved successfully');
```
dingo-d, line 50:
```suggestion
		$this->assertSame($response['success'], 'Candidate saved successfully');
```

Always use `assertSame`, because it will check the type and the value. You can fix this in the entire test suite.

dingo-d (changes requested):
Left a few comments

mbmjertan (dismissed):

tests/BaseTest.php
iruzevic: this is actually not ok, because the logic is reversed

tests/Integrations/Greenhouse/GreenhouseTest.php
iruzevic: what should I add in after?

tests/Integrations/Greenhouse/GreenhouseTest.php
iruzevic: 👍

dingo-d (dismissed):

tests/BaseTest.php
dingo-d: Then use `!==` instead 😄

tests/Integrations/Greenhouse/GreenhouseTest.php
dingo-d: I think `parent::_after();`. And before that you can just set the greenhouse property to null.

IvanGrginovInf: Parent class (`BaseTest`) already does `_after()` like so:

```php
	protected function _after()
	{
		Monkey\tearDown();
	}
```

No need to add `_after()` in your test unless you need to tearDown something else as well. CC @dingo-d @iruzevic

IvanGrginovInf (dismissed):

tests/BaseTest.php
iruzevic: Fixed

iruzevic:
Tnx guys

PR #96, opened 2021-10-04
Extra loading and error classes

goranalkovic-infinum:
Adds an extra `is-loading` class while fetching validation info etc. and a `has-error` to the whole field if it has an error, all to aid styling (some projects require input fields have a red border etc.)

### DIFF
--- a/src/Blocks/components/form/assets/form.js
+++ b/src/Blocks/components/form/assets/form.js
@@ -46,8 +46,7 @@ export class Form {
 				return response.json();
 			})
 			.then((response) => {
-				// Reset all error classes on fields.
-				element.querySelectorAll(`.${this.CLASS_HAS_ERROR}`).forEach((e) => e.classList.remove(this.CLASS_HAS_ERROR));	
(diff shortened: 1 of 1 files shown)


src/Blocks/components/form/assets/form.js
```diff
@@ -40,51 +42,52 @@ export class Form {
...
-						window.location.href = isRedirect;
-					}, 1000);
-				} else {
-					this.setGlobalMsg(element, response.message, 'success');
+			.then((response) => {
+				return response.json();
+			})
+			.then((response) => {
+				// Reset all error classes on fields.
+				element.querySelectorAll(`.${this.CLASS_HAS_ERROR}`).forEach((e) => e.classList.remove(this.CLASS_HAS_ERROR));
```
iruzevic, line 50: please make this as a separate method like hideLoader

iruzevic (changes requested):

iruzevic (approved):

PR #114, opened 2021-10-05
After submit URL redirect templating

goranalkovic-infinum:
Adds basic templating to "Redirect on submit success" URLs.

### DIFF
--- a/src/Blocks/components/form/assets/form.js
+++ b/src/Blocks/components/form/assets/form.js
@@ -55,13 +55,20 @@ export class Form {
 					this.gtmSubmit(element);
 
 					// If success, redirect or output msg.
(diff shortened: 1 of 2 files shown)


iruzevic:
Could you explain what this does exactly?

goranalkovic-infinum:
> Could you explain what this does exactly?

If you need to redirect after submitting the form somewhere with a specific URL - e.g (https://mysite.com/register?email=mymail@mail.com) that should have parts prefilled from the form data. In that case in the URL you would write `https://mysite.com/register?email={email}` and it would get replaced in code

iruzevic:
nice so any field name will be replaced nice

iruzevic (approved):

PR #119, opened 2021-10-06
adding validation

iruzevic:
- added validation fixes for global variables
- added msg for missing global configuration
- adding steps how to set up global settings

## PLEASE DON'T REVIEW HUBSPOT PART 
this is still WIP

Screenshot 2021-10-06 at 09 36 16

Screenshot 2021-10-06 at 09 36 09

### DIFF
--- a/src/Blocks/components/highlighted-content/highlighted-content-admin.scss
+++ b/src/Blocks/components/highlighted-content/highlighted-content-admin.scss
@@ -1,0 +1,22 @@
+.es-highlighted-content {
+	display: flex;
+	align-items: center;
+	justify-content: center;
+	min-height: 300px;
+	flex-direction: column;
+	text-align: center;
+
+	&__title {
+		font-size: 30px;
+		font-weight: bold;
+		color: var(--global-colors-es-boulder)
+	}
+
+	&__subtitle {
+		padding-top: 10px;
+		font-size: 16px;
+		line-height: 1.7;
+		color: var(--global-colors-es-boulder);
+		max-width: 50%;
+	}
+}
--- a/src/Rest/Routes/FormSubmitRoute.php
+++ b/src/Rest/Routes/FormSubmitRoute.php
@@ -157,6 +157,7 @@ class FormSubmitRoute extends AbstractBaseRoute
 			// Upload files to temp folder.
 			$files = $this->prepareFiles($files);
 
+			// Determin form type to use.
 			switch ($formType) {
 				case SettingsMailer::SETTINGS_TYPE_KEY:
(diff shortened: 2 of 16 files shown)


goranalkovic-infinum (approved):

iobrado (approved):

src/Rest/Routes/FormSubmitRoute.php
```diff
@@ -157,6 +157,7 @@ public function routeCallback(\WP_REST_Request $request)
 			// Upload files to temp folder.
 			$files = $this->prepareFiles($files);
 
+			// Determin form type to use.
```
volfkarlo, line 160:
```suggestion
			// Determine form type to use.
```

src/Rest/Routes/FormSubmitRoute.php
```diff
@@ -194,7 +195,7 @@ private function sendEmail(string $formId, array $params = [], $files = [])
 	{
 		$isUsed = (bool) $this->getSettingsValue(SettingsMailer::SETTINGS_MAILER_USE_KEY, $formId);
 
-		// If mailer system is not used just post as a success.
+		// If Mailer system is not used just post as a success.
```
volfkarlo, line 198:
```suggestion
		// If Mailer system is not used just respond with success.
```

volfkarlo (approved):
Micro corrections and a suggestion.

This is just a notice.
Did you make a check for already subscribed users?
On the Academy project we had problems with previously subscribed users. The problem arose when a user is created and then the same one wants to subscribe once more, you get an error message saying that the user is already created, that's why you have to update the user with the new mailing list, tag, or something similar to it.

src/Blocks/components/highlighted-content/highlighted-content-admin.scss
```diff
@@ -0,0 +1,22 @@
+.es-highlighted-content {
+	display: flex;
+	align-items: center;
+	justify-content: center;
+	min-height: 300px;
```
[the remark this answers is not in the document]
iruzevic: thx for the tip but this is WP admin settings and there is no responsive typography there so no need for this

iruzevic:
> Micro corrections and a suggestion.
> 
> This is just a notice. Did you make a check for already subscribed users? On the Academy project we had problems with previously subscribed users. The problem arose when a user is created and then the same one wants to subscribe once more, you get an error message saying that the user is already created, that's why you have to update the user with the new mailing list, tag, or something similar to it.

Please open an issue with this problem, and write in more detail what needs to happen when the user is all-ready signed to the mailing list. Tnx

PR #123, opened 2021-10-17
Feature/user confirmation

iruzevic:
# Worklog
- added sender-email block with pre filled options.
- added sender-name block with pre filled options.
- added changes to mails method and interface.
- added mailer options changes for user confirmation and addition changes for normal options.
- added submit route changes for user confirmation.

IMPORTANT: user confirmation email will only be sent if options are set and you have sender-email block added in the form

# Screenshot
Screenshot 2021-10-17 at 22 44 29
Screenshot 2021-10-17 at 22 44 32
Screenshot 2021-10-17 at 22 44 43

### DIFF
--- a/src/Blocks/custom/sender-email/sender-email.php
+++ b/src/Blocks/custom/sender-email/sender-email.php
@@ -1,0 +1,23 @@
+<?php
+
+/**
+ * Template for the Sender Email Block view.
+ *
+ * @package EightshiftForms
+ */
+
+use EightshiftForms\Helpers\Components;
+
+$unique = Components::getUnique();
+
+$inputName = $attributes['senderEmailInputName'] ?? '';
+$blockClass = $attributes['blockClass'] ?? '';
+
+$props['blockClass'] = $blockClass;
+
+echo Components::render( // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+	'input',
+	Components::props('input', $attributes, [
+		'inputId' => $inputName . $unique,
+	])
+);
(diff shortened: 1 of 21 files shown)


src/Blocks/custom/sender-email/sender-email.php
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+use EightshiftForms\Helpers\Components;
+
+$unique = Components::getUnique();
+
+$inputName = $attributes['senderEmailInputName'] ?? '';
+$blockClass = $attributes['blockClass'] ?? '';
+
+$props['blockClass'] = $blockClass;
```
volfkarlo, line 16: When is these `$props` variable used?

src/Blocks/custom/sender-name/sender-name.php
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+use EightshiftForms\Helpers\Components;
+
+$unique = Components::getUnique();
+
+$inputName = $attributes['senderNameInputName'] ?? '';
+$blockClass = $attributes['blockClass'] ?? '';
+
+$props['blockClass'] = $blockClass;
```
volfkarlo, line 16: same as above

iobrado (approved):

src/Blocks/custom/sender-email/sender-email.php
iruzevic: fixed

iobrado (approved):

PR #141, opened 2021-10-27
Validation fixes

iruzevic:
Refactored the way validation and passing data to the backed work because the previous version had a major bu bu


src/Blocks/components/checkbox/checkbox-style.scss
```diff
@@ -26,6 +26,14 @@
...
+		--es-checkbox-scoped-opacity: 0.7;
+	}
+
+	&__input:disabled:checked + #{$this}__label::after {
+		--es-checkbox-scoped-opacity: 0.7;
+	}
+
 	&__label {
 		position: relative;
 		padding-left: calc(var(--es-checkbox-scoped-icon-size) + 10px);
```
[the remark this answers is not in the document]
iruzevic: no need, we provide some simple styles for the forms but in 99% cases you will turn this off and provide your own styles so no need for the extra work here

src/Helpers/UploadHelper.php
```diff
@@ -41,18 +41,30 @@ protected function prepareFiles(array $files): array
...
+						'type' => $fileValue['type'][$key],
+						'tmp_name' => $fileValue['tmp_name'][$key],
+						'error' => $fileValue['error'][$key],
+						'size' => $fileValue['size'][$key],
+					];
+
+					$upload = wp_handle_upload(
+						$fileDetails,
+						[
+							'test_form' => false,
```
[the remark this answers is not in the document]
volfkarlo: wow, what a surprise 😂

src/Rest/Routes/FormSettingsSubmitRoute.php
```diff
@@ -83,17 +84,30 @@ public function routeCallback(\WP_REST_Request $request)
...
 		try {
+			$params = $this->prepareParams($request->get_body_params());
+
 			// Get encripted form ID and decrypt it.
-			$formId = $this->getFormId($request->get_body_params());
+			$formId = $this->getFormId($params);
 
-			// Determine form type.
-			$formType = $this->getFormType($request->get_body_params());
+			// Determin form type.
```
volfkarlo, line 92:
```suggestion
			// Determine form type.
```

volfkarlo (approved):
One typo! 😄

iobrado (approved):

src/Integrations/Mailchimp/SettingsMailchimp.php
```diff
@@ -264,7 +266,7 @@ public function getSettingsGlobalData(): array
 						'checkboxId' => $this->getSettingsName(self::SETTINGS_MAILCHIMP_USE_KEY),
 						'checkboxLabel' => __('Use Mailchimp', 'eightshift-forms'),
 						'checkboxIsRequired' => true,
-						'checkboxIsChecked' => !empty($this->getOptionValue(self::SETTINGS_MAILCHIMP_USE_KEY)),
+						'checkboxIsChecked' => $this->getOptionValue(self::SETTINGS_MAILCHIMP_USE_KEY) === 'true',
```
mbmjertan, line 269: Why is this a string?

mbmjertan (approved):

simun-ivanac (approved):

src/Blocks/components/radio/radio.php
```diff
@@ -42,11 +40,9 @@ class="<?php echo esc_attr("{$componentClass}__input"); ?>"
 			name="<?php echo esc_attr($radioName); ?>"
 			id="<?php echo esc_attr($radioId); ?>"
 			value="<?php echo esc_attr($radioValue); ?>"
-			data-validation-required="<?php echo esc_attr($radioIsRequired); ?>"
 			data-tracking="<?php echo esc_attr($radioTracking); ?>"
 			<?php echo $radioIsChecked ? 'checked' : ''; ?>
 			<?php echo $radioIsDisabled ? 'disabled' : ''; ?>
```
dingo-d, line 45: You can use https://developer.wordpress.org/reference/functions/checked/ and https://developer.wordpress.org/reference/functions/disabled/ for this 🙂

src/Form/AbstractFormBuilder.php
```diff
@@ -179,11 +179,17 @@ private function buildComponent(array $attributes): string
 					break;
 			}
 
+			$id = $attributes["{$component}Id"] ?? '';
+
 			// Loop children and do the same ad on top level.
-			foreach ($attributes[$key] as $item) {
+			foreach ($attributes[$key] as $innerKey => $item) {
 				// Determin component name.
```
dingo-d, line 186:
```suggestion
				// Determine the component's name.
```

src/Form/AbstractFormBuilder.php
```diff
@@ -179,11 +179,17 @@ private function buildComponent(array $attributes): string
...
 				$innercComponent = $item['component'] ? HelpersComponents::kebabToCamelCase($item['component']) : '';
 
+				if ($key === 'radiosContent') {
+					$item["{$innercComponent}Id"] = "{$id}[{$innerKey}]";
```
dingo-d, line 190: The `$id` will fallback to empty string in case the `$attributes["{$component}Id"]` is not set or empty. In this case, the right-hand side will throw an error since you're trying to get a key out of a string.

Or am I missing something?

src/Helpers/UploadHelper.php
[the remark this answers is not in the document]
dingo-d: Jup, if you dig deeper: https://developer.wordpress.org/reference/functions/_wp_handle_upload/

WordPress is... interesting 😂

src/Integrations/Greenhouse/Greenhouse.php
```diff
@@ -74,23 +84,35 @@ public function getForm(array $formAdditionalProps): string
...
-			return '';
+			return [];
 		}
 
-		// Return form to the frontend.
-		return $this->buildForm(
-			$this->getFields($questions, (string) $formId),
-			$formAdditionalProps
-		);
+		return $this->getFields($questions, (string) $formId);
```
dingo-d, line 115: Is the `$formId` some numerical value or some string identifier? Like `form-12345`? Or just `123456`?

src/Integrations/Mailchimp/SettingsMailchimp.php
dingo-d: I'd recon that the `getOptionValue` method returns a string. I'd guess it can return more than just true/false values. Like keys and other things.

src/Settings/GlobalSettings/SettingsGlobal.php
```diff
@@ -138,10 +123,12 @@ private function getAllSettingsSidebars(): array
...
 	{
-		$allSettings = self::GLOBAL_SETTINGS;
+		$allSettings = [];
 
-		foreach (Integrations::ALL_INTEGRATIONS as $key => $integration) {
-			$allSettings[$key] = $integration['global'] ?? '';
+		foreach (Filters::ALL as $key => $integration) {
+			if (isset($integration['global'])) {
+				$allSettings[$key] = $integration['global'] ?? '';
+			}
```
dingo-d, line 131: Shouldn't `$integration['global'] ?? ''` be enough (unnecessary check for isset)?

src/Settings/Settings/SettingsAll.php
```diff
@@ -179,10 +162,12 @@ private function getAllSettingsSidebars(): array
...
 	{
-		$allSettings = self::ALL_SETTINGS;
+		$allSettings = [];
 
-		foreach (Integrations::ALL_INTEGRATIONS as $key => $integration) {
-			$allSettings[$key] = $integration['settings'] ?? '';
+		foreach (Filters::ALL as $key => $integration) {
+			if (isset($integration['settings'])) {
+				$allSettings[$key] = $integration['settings'] ?? '';
+			}
```
dingo-d, line 170: Same as the above comment. Or if you don't need to set it to empty, you can just use `continue` inside the foreach to skip the integrations that aren't set.

src/Settings/Settings/SettingsTest.php
```diff
@@ -0,0 +1,293 @@
...
+						'component' => 'select-option',
+						'selectOptionLabel' => 'select-label2',
+						'selectOptionValue' => 'select-id2',
+						'selectOptionIsSelected' => $this->getOptionValue('select-required-id') === 'select-id2',
+					],
+				]
+			],
+			[
+				'component' => 'checkboxes',
+				'checkboxesFieldLabel' => \__('checkboxes', 'eightshift-forms'),
```
dingo-d, line 203: Global prefix on translation function looks odd.

src/Validation/Validator.php
```diff
@@ -95,50 +137,224 @@ private function validateParams(array $params, string $formId): array
...
-					case 'validationMinSize':
-						if (!empty($dataValue) && !$this->isFileMinSizeValid((int) $fileSize, (int) $dataValue * 1000) && $inputData['validationRequired'] === '1') {
-							$output[$inputName] = sprintf($this->labels->getLabel('validationMinSize', $formId), $dataValue);
+					}
+				}
+
+				// Check validation for size min/max.
+				foreach ($fileValue['size'] as $file) {
+					// Check validation for min size.
+					if ($dataKey === 'minSize') {
```
dingo-d, line 177: This looks a bit odd. You're checking the key of the previous loop, inside another loop. 

Not sure I understand the flow here.

src/Validation/Validator.php
```diff
@@ -95,50 +137,224 @@ private function validateParams(array $params, string $formId): array
...
+						if (!empty($dataValue) && !$this->isFileMinSizeValid((int) $file, (int) $dataValue * 1000)) {
```
dingo-d, line 178: What can the `$file` be? Something other than integer? Because you're checking the size, but pass the file. Naming is confusing me.

dingo-d (changes requested):
Left a few comments and suggestions. Great job 🙂

src/Rest/Routes/FormSubmitRoute.php
```diff
@@ -139,23 +140,27 @@ public function routeCallback(\WP_REST_Request $request)
 
 		// Try catch request.
 		try {
+			$params = $this->prepareParams($request->get_body_params());
+
 			// Get encripted form ID and decrypt it.
```
dadadavorin, line 145: // Get encrypted ...

src/Rest/Routes/FormSubmitRoute.php
```diff
@@ -139,23 +140,27 @@ public function routeCallback(\WP_REST_Request $request)
...
-			$formId = $this->getFormId($request->get_body_params(), true);
+			$formId = $this->getFormId($params, true);
 
 			// Determin form type.
```
dadadavorin, line 148: // Determine ...

dadadavorin (approved):
Wow 🤩  Awesome work here ! 👍

src/Blocks/components/radio/radio.php
iruzevic: tnx

src/Form/AbstractFormBuilder.php
iruzevic: I see the confusion but this will not actually get any thing but rather output this kind of string blablaSomeID[0], blablaSomeID[1], etc for the radio name attribute something like this https://stackoverflow.com/questions/11624871/radiobutton-array-html

src/Integrations/Greenhouse/Greenhouse.php
iruzevic: as I see it is numerical only but I cant be sure so in whole project I used string for all formId identifiers.

src/Integrations/Mailchimp/SettingsMailchimp.php
iruzevic: if the checkbox value is set it will be used if the checkbox is checked. If it is not checked it will return false. I'm trying to reduce type jugling here so I cast all in string. Because bool can be set in the db as true and returned as true but outputed as 1 and so on. In this way sa string I'm safe (ish)

src/Settings/GlobalSettings/SettingsGlobal.php
iruzevic: not sure but it looks to me that it will pass the empty string to the array this way. 
I can refactor this for a return early method if you would like

src/Settings/Settings/SettingsTest.php
iruzevic: fixed in the other PR

src/Validation/Validator.php
iruzevic: yea the problem is in the the way you get the array of values in the multiple file. You will get something like this:
```php

[02-Nov-2021 13:34:54 UTC] Array
(
    [898b0eb2e944aece5e2ece26e5519d12] => Array
        (
            [name] => Array
                (
                    [0] => Image from iOS (1).jpg
                    [1] => Image from iOS (2).jpg
                    [2] => Image from iOS (3).jpg
                    [3] => Image from iOS (4).jpg
                )

            [type] => Array
                (
                    [0] => image/jpeg
                    [1] => image/jpeg
                    [2] => image/jpeg
                    [3] => image/jpeg
                )

            [tmp_name] => Array
                (
                    [0] => /private/var/tmp/phpIESHaY
                    [1] => /private/var/tmp/phpS2jJBm
                    [2] => /private/var/tmp/phpoCGnQi
                    [3] => /private/var/tmp/phpc2hX7F
                )

            [error] => Array
                (
                    [0] => 0
                    [1] => 0
                    [2] => 0
                    [3] => 0
                )

            [size] => Array
                (
                    [0] => 39637
                    [1] => 36483
                    [2] => 17234
                    [3] => 159145
                )

        )

)
```

So every key in the array reppresents one file this is why you need to check for the key in the parent. Stupid I know

src/Validation/Validator.php
iruzevic: fixed naming

src/Form/AbstractFormBuilder.php
dingo-d: Oh! I see, I omitted the `{}` around the ID. Got it 👍🏼

src/Integrations/Greenhouse/Greenhouse.php
```diff
@@ -74,23 +84,35 @@ public function getForm(array $formAdditionalProps): string
 			return '';
 		}
 
+		// Return form to the frontend.
+		return $this->buildForm(
+			$this->getFormFields((string) $formId),
```
dingo-d, line 89: You're casting this to string on line 82, so I don't think you need to do that here. Or just remove it on line 82, so this value will always be a string.

Come to think of it, remove it on line 82, as there you just casted the first expression of the ternary to string, then got something as a result that didn't have to be a string

https://3v4l.org/jC0XR

src/Settings/GlobalSettings/SettingsGlobal.php
dingo-d: Sure, but my point is that the value of `$allSettings[$key]` will either be `$integration['global']` or an empty string,  if the `$integration['global']` is not empty and set. So the check above isn't necessary.

Or if you want the conditional to be necessary, then the `?? ''` part is not necessary.

src/Validation/Validator.php
dingo-d: butwait

dingo-d (approved):

dingo-d (approved):

PR #175, opened 2021-11-12
refactoring storage to custom db

iruzevic:
# Worklog
- Adding custom db setup for storing information

### DIFF
--- a/src/Settings/SettingsHelper.php
+++ b/src/Settings/SettingsHelper.php
@@ -10,9 +10,9 @@ declare(strict_types=1);
 
 namespace EightshiftForms\Settings;
 
+use EightshiftForms\Config\Config;
 use EightshiftForms\Helpers\Components;
 use EightshiftForms\Integrations\Greenhouse\SettingsGreenhouse;
-use EightshiftForms\Integrations\Mailchimp\Mailchimp;
 
 /**
  * SettingsHelper trait.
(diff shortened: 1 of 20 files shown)


src/Settings/SettingsHelper.php
```diff
@@ -354,4 +548,42 @@ public function getIntegrationFieldsValue(array $dbSettingsValue, array $formFie
...
+		global $wpdb;
+
+		$tableNameSettings = "{$wpdb->prefix}" . Config::getDbSettingsName();
+		$tableNameOptions = "{$wpdb->prefix}" . Config::getDbOptionsName();
+
+		$sqlSettings = "CREATE TABLE {$tableNameSettings} ( 
+			meta_id BIGINT(20) UNSIGNED NOT NULL AUTO_INCREMENT,
+			post_id BIGINT(20) UNSIGNED NOT NULL DEFAULT '0',
+			meta_key VARCHAR(255) NULL DEFAULT NULL,
+			meta_value LONGTEXT NULL DEFAULT NULL,
```
dingo-d, line 568: What kind of data is going to be stored here? Serialized arrays, JSON, something else?

iruzevic: basicly any kind. It can be string or Serialized arrays, JSON etc

src/Settings/SettingsHelper.php
```diff
@@ -354,4 +548,42 @@ public function getIntegrationFieldsValue(array $dbSettingsValue, array $formFie
...
+			PRIMARY KEY (meta_id),
+			INDEX post_id (post_id),
+			INDEX meta_key (meta_key)
```
mbmjertan, line 571: Will it be necessary to query based on `meta_value` or a `meta_key` and `meta_value` combination? If so, I'd suggest adding an index on `meta_value` in the former case, that is, on `meta_key+meta_value` in the latter case. A good read regarding such indices is the WPVIP docs on the WordPress `postmeta` table: https://docs.wpvip.com/technical-references/code-quality-and-best-practices/querying-on-meta_value/

src/Settings/SettingsHelper.php
dingo-d: Yeah, I'm not too happy about that. Custom DB tables should be custom following all the best practices (normal forms etc.) of database design.

Just copy-pasting meta and option tables make little sense to me tbh, plus you're losing all the caching WP offers.

iruzevic:
this is dropped

PR #193, opened 2021-12-07
Update frontend libs

goranalkovic-infinum:


goranalkovic-infinum:
@iruzevic
Looks like the package file was upgraded to v2, which we may not want

iruzevic (approved):

PR #218, opened 2022-03-15
HubSpot integration fixes and improvements

mbmjertan:
# Description

This PR:
* changes the `HubspotClient::getApiKey` method conditional so we return API keys properly
* changes the `HubspotClient::getItems` method to return `submitText` as well
* changes the `Hubspot::getFormFields` and `getFields` methods to consume the whole `$item` array so `submitText` can be used (without mapping it to a field manually in `HubspotClient`)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,6 +64,8 @@ jobs:
           php-version: ${{ matrix.php }}
           coverage: none
           tools: cs2pr
+        env:
+          COMPOSER_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       # Install dependencies and handle caching in one go.
       # @link https://github.com/marketplace/actions/install-composer-dependencies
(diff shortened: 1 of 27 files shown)


iruzevic (changes requested):
I need to check this before the merge I'm not sure in this implementation

.github/workflows/ci.yml
```diff
@@ -13,76 +13,103 @@ on:
     paths-ignore:
       - '**.md'
 
-# Cancels all previous workflow runs for the same branch that have not yet completed.
```
dingo-d, line 16: Leave the concurrency in 🙏🏼

iruzevic (approved):
I have tested this and it is ok. tnx

.github/workflows/ci.yml
mbmjertan: @dingo-d, sure, but I don't know how to fix these CI issues. I can remove these changes entirely, this was just testing to see how to fix CI.

Unfortunately this is still stuck on Composer failing to authenticate against GitHub.

dingo-d:
@mbmjertan I've updated some composer packages, so phpstan will probably fail. Also, I may have stumbled upon another edge case for the coding standards so I'll have to fix those, which is why I constrained the standards to v1.3.

Should be ok now.

dingo-d (approved):

PR #220, opened 2022-03-25
Feature/clearbit

iruzevic:
# Description

- new integration with Clearbit.
- updated style for field settings for integrations
- Integration icon on the listing if every thing is set correctly.
- Tooltips

# Screenshots / Videos
Screenshot 2022-03-25 at 10 45 00
Screenshot 2022-03-25 at 10 44 29
Screenshot 2022-03-25 at 10 44 39
Screenshot 2022-03-25 at 10 44 57

### DIFF
--- a/src/Integrations/Clearbit/ClearbitClient.php
+++ b/src/Integrations/Clearbit/ClearbitClient.php
@@ -31,6 +31,13 @@ class ClearbitClient implements ClearbitClientInterface
 	use ObjectHelperTrait;
 
 	/**
+	 * Return Clearbit base url.
+	 *
+	 * @var string
+	 */
+	private const BASE_URL = 'https://person-stream.clearbit.com/v2/';
+
+	/**
 	 * API request to post application.
 	 *
 	 * @param string $emailKey Email key to map in params.
@@ -78,7 +85,7 @@ class ClearbitClient implements ClearbitClientInterface
 		}
 
 		$response = \wp_remote_get(
-			"{$this->getBaseUrl()}combined/find?email={$email}",
+			self::BASE_URL . "combined/find?email={$email}",
 			[
 				'headers' => $this->getHeaders(),
 			]
@@ -302,16 +309,6 @@ class ClearbitClient implements ClearbitClientInterface
 	{
 		$apiKey = Variables::getApiKeyClearbit();
 
-		return $apiKey ? $apiKey : $this->getOptionValue(SettingsClearbit::SETTINGS_CLEARBIT_API_KEY_KEY); // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.obfuscation_base64_encode
-	}
-
-	/**
-	 * Return Clearbit base url.
-	 *
-	 * @return string
-	 */
-	private function getBaseUrl(): string
-	{
-		return 'https://person-stream.clearbit.com/v2/';
+		return $apiKey ? $apiKey : $this->getOptionValue(SettingsClearbit::SETTINGS_CLEARBIT_API_KEY_KEY);
 	}
 }
--- a/src/Integrations/Clearbit/ClearbitClient.php
+++ b/src/Integrations/Clearbit/ClearbitClient.php
@@ -12,14 +12,13 @@ namespace EightshiftForms\Integrations\Clearbit;
 
 use EightshiftForms\Helpers\Helper;
 use EightshiftForms\Hooks\Variables;
-use EightshiftForms\Integrations\ClientInterface;
 use EightshiftForms\Settings\SettingsHelper;
 use EightshiftFormsVendor\EightshiftLibs\Helpers\ObjectHelperTrait;
 
 /**
  * ClearbitClient integration class.
  */
-class ClearbitClient implements ClientInterface
+class ClearbitClient implements ClearbitClientInterface
 {
 	/**
 	 * Use general helper trait.
@@ -32,54 +31,56 @@ class ClearbitClient implements ClientInterface
 	use ObjectHelperTrait;
 
 	/**
-	 * Return items.
-	 *
-	 * @param bool $hideUpdateTime Determin if update time will be in the output or not.
-	 *
-	 * @return array<string, mixed>
-	 */
-	public function getItems(bool $hideUpdateTime = true): array
-	{
-		$output = [];
-
-		foreach ($this->prepareParams() as $key => $value) {
-			$output[] = $key;
-		}
-
-		return $output;
-	}
-
-	/**
-	 * Return item with cache option for faster loading.
-	 *
-	 * @param string $itemId Item ID to search by.
-	 *
-	 * @return array<string, mixed>
-	 */
-	public function getItem(string $itemId): array
-	{
-		return [];
-	}
-
-	/**
 	 * API request to post application.
 	 *
-	 * @param string $itemId Item id to search.
+	 * @param string $emailKey Email key to map in params.
 	 * @param array<string, mixed> $params Params array.
-	 * @param array<string, array<int, array<string, mixed>>> $files Files array.
-	 * @param string $formId FormId value.
+	 * @param array<string, string> $mapData Map data from settings.
 	 *
 	 * @return array<string, mixed>
 	 */
-	public function postApplication(string $itemId, array $params, array $files, string $formId): array
+	public function getApplication(string $emailKey, array $params, array $mapData): array
 	{
+		$email = isset($params[$emailKey]['value']) ? $params[$emailKey]['value'] : '';
 
-		$email = isset($params[$itemId]) ? $params[$itemId]['value'] : '';
+		if (!$email) {
+			$output = [
+				'status' => 'error',
+				'code' => 400,
+				'message' => 'clearbitMissingEmail',
+			];
+
+			Helper::logger([
+				'integration' => 'clearbit',
+				'email' => $email,
+				'mapKeys' => $mapData,
+				'output' => $output,
+			]);
+
+			return $output;
+		}
+
+		if (!$mapData) {
+			$output = [
+				'status' => 'error',
+				'code' => 400,
+				'message' => 'clearbitMissingMapKeys',
+			];
+
+			Helper::logger([
+				'integration' => 'clearbit',
+				'email' => $email,
+				'mapKeys' => $mapData,
+				'output' => $output,
+			]);
+
+			return $output;
+		}
 
 		$response = \wp_remote_get(
 			"{$this->getBaseUrl()}combined/find?email={$email}",
 			[
-				'headers' => $this->getHeaders(true),
+				'headers' => $this->getHeaders(),
 			]
 		);
 
@@ -96,11 +97,20 @@ class ClearbitClient implements ClientInterface
 		$responseBody = json_decode(\wp_remote_retrieve_body($response), true);
 
 		if ($code === 200) {
+			$dataOutput = [];
+
+			foreach ($this->prepareParams($responseBody) as $key => $value) {
+				if (array_key_exists($key, $mapData) && !empty($value)) {
+					$dataOutput[$mapData[$key]] = $value;
+				}
+			}
+
 			return [
 				'status' => 'success',
 				'code' => $code,
 				'message' => 'clearbitSuccess',
-				'data' => $this->prepareParams($responseBody),
+				'email' => $email,
+				'data' => $dataOutput,
 			];
 		}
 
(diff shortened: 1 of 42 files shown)


src/Integrations/Clearbit/ClearbitClient.php
```diff
@@ -0,0 +1,317 @@
...
+	private function getHeaders(): array
+	{
+		return [
+			'Content-Type' => 'application/json; charset=utf-8',
+			'Authorization' => "Bearer {$this->getApiKey()}",
+		];
+	}
+
+	/**
+	 * Return Api Key from settings or global vairaible.
```
iobrado, line 297:
```suggestion
	 * Return Api Key from settings or global variable.
```

iobrado (approved):

gabriel-glo (approved):

goranalkovic-infinum (approved):

simun-ivanac (approved):

volfkarlo (approved):
Seems great! 😄 

Just a question, is there some kind of delay between Hubspot submission and Clearbit update to the contact?

I remember that we had a Clearbit update scheduled on the cron job due to potential hickups on Hubspot.

(This info is needed for workflows in Hubspot)

iruzevic:
> Seems great! 😄
> 
> Just a question, is there some kind of delay between Hubspot submission and Clearbit update to the contact?
> 
> I remember that we had a Clearbit update scheduled on the cron job due to potential hickups on Hubspot.
> 
> (This info is needed for workflows in Hubspot)

There is no need for this anymore, I have found a solution

src/Integrations/Clearbit/ClearbitClient.php
```diff
@@ -0,0 +1,317 @@
...
+	/**
+	 * Return Api Key from settings or global variable.
+	 *
+	 * @return string
+	 */
+	private function getApiKey(): string
+	{
+		$apiKey = Variables::getApiKeyClearbit();
+
+		return $apiKey ? $apiKey : $this->getOptionValue(SettingsClearbit::SETTINGS_CLEARBIT_API_KEY_KEY); // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.obfuscation_base64_encode
```
dingo-d, line 305: Why do you have the ignore phpcs comment here? I don't see base64_encode used here 🤷🏼‍♂️

src/Integrations/Clearbit/ClearbitClient.php
```diff
@@ -0,0 +1,317 @@
...
+	}
+
+	/**
+	 * Return Clearbit base url.
+	 *
+	 * @return string
+	 */
+	private function getBaseUrl(): string
```
dingo-d, line 313: This method is only used once in the class and is returning the URI string. I'd put this as a `private const` instead.

src/Integrations/Clearbit/SettingsClearbit.php
```diff
@@ -0,0 +1,456 @@
...
+
+	/**
+	 * Determine if settings are valid.
+	 *
+	 * @param string $formId Form ID.
+	 * @param string $type Integration type.
+	 *
+	 * @return boolean
+	 */
+	public function isSettingsValid(string $formId, string $type): bool
```
dingo-d, line 106:
```suggestion
	public function areSettingsValid(string $formId, string $type): bool
```

You are checking multiple settings, no? If that's the case, plural seems more fitting here.

src/Integrations/Clearbit/SettingsClearbit.php
```diff
@@ -0,0 +1,456 @@
...
+
+		return true;
+	}
+
+	/**
+	 * Determine if settings global are valid.
+	 *
+	 * @return boolean
+	 */
+	public function isSettingsGlobalValid(): bool
```
dingo-d, line 144: Same here about the plural here. `areSettingsGlobalValid` reads better

src/Integrations/Clearbit/SettingsClearbit.php
```diff
@@ -0,0 +1,456 @@
...
+			[
+				'component' => 'intro',
+				'introTitle' => __('Clearbit', 'eightshift-forms'),
+			],
+			[
+				'component' => 'intro',
+				'introTitle' => __('How to get the API key?', 'eightshift-forms'),
+				'introTitleSize' => 'small',
+				// phpcs:ignore WordPress.WP.I18n.NoHtmlWrappedStrings
+				'introSubtitle' => __('<ol>
```
dingo-d, line 201: I'm not overly fond of seeing HTML in a translatable string. `sprintf` would be a better match here IMO. That way, you can have HTML outside of the translatable strings.

src/Integrations/Clearbit/SettingsClearbit.php
```diff
@@ -0,0 +1,456 @@
...
+					),
+				],
+			];
+		}
+
+		return array_merge(
+			$output,
+			$outputApi,
+			$outputUsage
+		);
```
dingo-d, line 282: Would `$output + $outputApi + $outputUsage` return the same array?

src/Integrations/Clearbit/SettingsClearbit.php
```diff
@@ -0,0 +1,456 @@
...
+					'selectIsRequired' => true,
+					'selectValue' => $this->getSettingsValue($emailFieldKey, $formId),
+				],
+			];
+		}
+
+		return array_merge(
+			$output,
+			$outputEmail
+		);
```
dingo-d, line 375: Same as above comment

src/Integrations/Clearbit/SettingsClearbit.php
```diff
@@ -0,0 +1,456 @@
...
+	 * Output array settings for form.
+	 *
+	 * @param array<string, string> $properties Array of properties from integration.
+	 * @param array<string, string> $keys Array of keys to get data from.
+	 *
+	 * @return array<int, array<string, array<int|string, array<string, mixed>>|bool|string>>
+	 */
+	public function getOutputGlobalClearbit(array $properties, array $keys): array
+	{
+		$mapKey = isset($keys['map']) ? $keys['map'] : '';
```
dingo-d, line 388:
```suggestion
		$mapKey = $keys['map'] ?? '';
```

src/Integrations/Clearbit/SettingsClearbit.php
```diff
@@ -0,0 +1,456 @@
...
+			$clearbitAvailableKeys ? [
+				'component' => 'group',
+				'groupId' => $this->getSettingsName($mapKey),
+				'groupStyle' => 'default',
+				'groupContent' => [
+					[
+						'component' => 'group',
+						'groupSaveOneField' => true,
+						'groupContent' =>  array_map(
+							function ($item) use ($clearbitMapValue, $properties) {
```
dingo-d, line 419:
```suggestion
							static function ($item) use ($clearbitMapValue, $properties) {
```

src/Integrations/Clearbit/SettingsClearbit.php
```diff
@@ -0,0 +1,456 @@
...
+									'selectOptions' => array_merge(
+										[
+											[
+												'component' => 'select-option',
+												'selectOptionLabel' => '',
+												'selectOptionValue' => '',
+											],
+										],
+										array_map(
+											function ($option) use ($selectedValue) {
```
dingo-d, line 436:
```suggestion
											static function ($option) use ($selectedValue) {
```

src/Integrations/Greenhouse/SettingsGreenhouse.php
```diff
@@ -324,7 +325,7 @@ public function getSettingsGlobalData(): array
 			[
 				'component' => 'intro',
 				'introTitle' => __('How to get the Job Board name?', 'eightshift-forms'),
-				'introTitleSize' => 'medium',
+				'introTitleSize' => 'small',
 				// phpcs:ignore WordPress.WP.I18n.NoHtmlWrappedStrings
 				'introSubtitle' => __('<ol>
```
dingo-d, line 330: Translation using `sprintf` to avoid placing HTML in translatable string.

src/Integrations/Hubspot/HubspotClient.php
```diff
@@ -114,6 +119,52 @@ public function getItem(string $itemId): array
...
+
+		if (empty($output)) {
+			$items = $this->getHubspotContactProperties();
+
+			$output = [];
+
+			$allowedTypes = array_flip([
+				'text',
+				'textarea',
+			]);
```
dingo-d, line 139: This is a manual array, why not just writing

```suggestion
			$allowedTypes = [
				'text' => 0,
				'textarea' => 1,
			];
```

?

src/Integrations/Hubspot/SettingsHubspot.php
```diff
@@ -183,6 +220,7 @@ public function getSettingsSidebar(): array
 	 */
 	public function getSettingsData(string $formId): array
 	{
+		// Bailout if global config is not valid.
 		if (!$this->isSettingsGlobalValid()) {
```
dingo-d, line 224: `areSettingsGlobalValid` sounds better. Like one of the above comment.

src/Integrations/Hubspot/SettingsHubspot.php
```diff
@@ -379,11 +339,150 @@ public function getSettingsGlobalData(): array
...
+
+		return [
+			[
+				'component' => 'divider',
+			],
+			[
+				'component' => 'intro',
+				'introTitle' => __('Form fields', 'eightshift-forms'),
+				'introTitleSize' => 'medium',
+				'introSubtitle' => __('Control which fields show up on the frontend, set up how they look and work.', 'eightshift-forms'),
```
dingo-d, line 472:
```suggestion
				'introSubtitle' => __('Controls which fields show up on the frontend, set up how they look and work.', 'eightshift-forms'),
```

dingo-d (changes requested):
Great work, I've left some comments in the PR 👍🏼

mbmjertan (approved):

src/Integrations/Clearbit/ClearbitClient.php
iruzevic: copy from the other file

src/Integrations/Clearbit/ClearbitClient.php
iruzevic: i would like to have consistent client files for all integrations. In hubspot I have multiple Urls so I would like to leave this as is

src/Integrations/Clearbit/SettingsClearbit.php
iruzevic: this is consistent like this across all integrations and settings so i would like to leave it as is

src/Integrations/Clearbit/SettingsClearbit.php
iruzevic: Not sure but I would like to leave it as is now

src/Integrations/Greenhouse/SettingsGreenhouse.php
iruzevic: this will be refactored in the next release so I would leave it as is for now

src/Integrations/Clearbit/ClearbitClient.php
dingo-d: Sure, but this is a private method that is different than the other one (the other ones have arguments, this one does not). So they're not that consistent 😅 

If you see certain parts repeating across files you can always abstract it out in an abstract class like a template pattern.

src/Integrations/Clearbit/SettingsClearbit.php
dingo-d: Since the eightshift forms are still in their alpha phase, you can just rename all of those method names. You're not breaking anything by doing that.

src/Integrations/Clearbit/SettingsClearbit.php
dingo-d: Sure, can you test this afterward?

dingo-d:
Can you fix the phpstan and phpcs issues?

src/Integrations/Clearbit/SettingsClearbit.php
iruzevic: yes but we have on some places one setting that needs to be valid and on some multiple settings, but in the usage I don't care about that, I just care that it is valid. So I want to have a consistant method name for valid.

I will leave it as is

dingo-d (approved):

PR #222, opened 2022-04-08
Test coverage improvements and bugfixes

mbmjertan:
# Description

This PR improves test coverage for the Helper and Validator class, fixes the Hubspot Settings test, and fixes some bugs in helper and validation methods:
* fixes typo in `convertInnerBlocksToArray` name
* changes the regex in `convertInnerBlocksToArray` to match more cases
* adds a typehint to `logger` to prevent undefined index errors
* fixes regex pattern validation

I'll probably do more work here, but feedback is really appreciated.

### DIFF
--- a/src/Validation/AbstractValidation.php
+++ b/src/Validation/AbstractValidation.php
@@ -95,6 +95,10 @@ abstract class AbstractValidation implements ValidatorInterface
 	public function isMimeTypeValid(array $file): bool
 	{
 		$denyIfFileIsNotUploaded = \apply_filters(Filters::getValidationSettingsFilterName('failMimetypeValidationWhenFileNotOnFS'), false); // phpcs:ignore WordPress.NamingConventions.ValidHookName.NotLowercase
+		if (getenv('TEST')) {
+			$denyIfFileIsNotUploaded = getenv('test_force_option_eightshift_forms_force_mimetype_from_fs') ?? false;
+		}
+
 		$mimeTypes = \array_flip(\wp_get_mime_types());
 
 		$fileMimetype = $file['type'];
--- a/src/Validation/AbstractValidation.php
+++ b/src/Validation/AbstractValidation.php
@@ -99,7 +99,14 @@ abstract class AbstractValidation implements ValidatorInterface
 
 		$fileMimetype = $file['type'];
 		if ($file['tmp_name'] ?? false) {
-			$fileMimetype = \mime_content_type($file['tmp_name']);
+			try {
+				$fileMimetype = \mime_content_type($file['tmp_name']);
+			}
+			catch (\Throwable $t) {
+				if ($denyIfFileIsNotUploaded) {
+					return false;
+				}
+			}
 		} elseif ($denyIfFileIsNotUploaded) {
 			return false;
 		}
--- a/src/Validation/AbstractValidation.php
+++ b/src/Validation/AbstractValidation.php
@@ -101,8 +101,7 @@ abstract class AbstractValidation implements ValidatorInterface
 		if ($file['tmp_name'] ?? false) {
 			try {
 				$fileMimetype = \mime_content_type($file['tmp_name']);
-			}
-			catch (\Throwable $t) {
+			} catch (\Throwable $t) {
 				if ($denyIfFileIsNotUploaded) {
 					return false;
 				}
--- a/src/Validation/AbstractValidation.php
+++ b/src/Validation/AbstractValidation.php
@@ -10,6 +10,8 @@ declare(strict_types=1);
 
 namespace EightshiftForms\Validation;
 
+use EightshiftForms\Hooks\Filters;
+
 /**
  * Class Validation
  */
@@ -75,9 +77,62 @@ abstract class AbstractValidation implements ValidatorInterface
 	 */
 	public function isFileTypeValid(string $fileName, string $fileTypes): bool
 	{
-		$fileExtension = explode('.', $fileName);
-		$validTypes = explode(',', str_replace(' ', '', str_replace('.', '', $fileTypes)));
+		$validTypes = $this->parseFiletypesString($fileTypes);
 
-		return in_array(end($fileExtension), $validTypes, true);
+		return in_array($this->getFileExtensionFromFilename($fileName), $validTypes, true);
+	}
+
+	/**
+	 * Checks whether the mimetype for the file is valid,
+	 * i.e. that it matches the extension. If the file is written
+	 * to disk, it'll check its mime_content_type from the filesystem.
+	 * Use the validation filter failMimetypeValidationWhenFileNotOnFS
+	 * to override this behaviour.
+	 * 
+	 * @param array $file File array.
+	 * @return boolean True if mimetype matches extension, false otherwise.
+	 */
+	public function isMimeTypeValid(array $file): bool {
+		$denyIfFileIsNotUploaded = apply_filters(Filters::getValidationSettingsFilterName('failMimetypeValidationWhenFileNotOnFS'), false); 
+		$mimeTypes = \array_flip(\wp_get_mime_types());
+
+		$fileMimetype = $file['type'];
+		if ($file['tmp_name'] ?? false) {
+			$fileMimetype = \mime_content_type($file['tmp_name']);
+		}
+		elseif ($denyIfFileIsNotUploaded) {
+			return false;
+		}
+
+		$fileExtension = $this->getFileExtensionFromFilename($file['name']);
+
+		if (in_array($fileExtension, explode('|', $mimeTypes[$fileMimetype] ?? []))) {
+			return true;
+		}
+
+		return false;
+	}
+
+	/**
+	 * Parses a comma-separated list of file extensions to return
+	 * a normalized array of allowed extenstions.
+	 * 
+	 * @param string $fileTypes String of file extensions, e.g. ".pdf, jpg,.gif"
+	 * @return array Array of extensions, e.g. ["pdf", "jpg", "gif"]
+	 */
+	private function parseFiletypesString(string $fileTypes): array {
+		$validTypes = explode(',', str_replace(' ', '', str_replace('.', '', $fileTypes)));
+		return $validTypes;
+	}
+
+	/**
+	 * Given a filename, returns its extension.
+	 *
+	 * @param string $fileName File name.
+	 * @return string Extension or filename if no extension.
+	 */
+	private function getFileExtensionFromFilename(string $fileName): string {
+		$explodedFilename = explode('.', $fileName);
+		return end($explodedFilename);
 	}
 }
--- a/src/Validation/AbstractValidation.php
+++ b/src/Validation/AbstractValidation.php
@@ -24,7 +24,7 @@ abstract class AbstractValidation implements ValidatorInterface
 	 */
 	public function isUrl(string $string): bool
 	{
-		return (bool) preg_match('/(http|ftp|mailto)/', $string);
+		return (bool) preg_match('/(http:\/\/|https:\/\/|ftp:\/\/|mailto:)/', $string);
 	}
 
 	/**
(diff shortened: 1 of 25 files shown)


tests/Helpers/HelperTest.php
```diff
@@ -60,3 +61,123 @@
...
+
+//---------------------------------------------------------------------------------//
+
+test('getFormTrashActionUrl calls get_delete_post_link with correct arguments.', function () {
+	$this->assertSame('id: 0, force: false', Helper::getFormTrashActionUrl('0'));
+	$this->assertSame('id: 0, force: false', Helper::getFormTrashActionUrl('0', 0));
+	$this->assertSame('id: 0, force: true', Helper::getFormTrashActionUrl('0', true));
+	$this->assertSame('id: 0, force: true', Helper::getFormTrashActionUrl('0', 1));
+	$this->assertSame('id: 75, force: true', Helper::getFormTrashActionUrl('75', 1));
+});
```
dingo-d, line 73: You could use expectations API coupled with a datasets for similar returning cases.

https://pestphp.com/docs/datasets
https://pestphp.com/docs/expectations

The test would then be something like

```php
test('getFormTrashActionUrl calls get_delete_post_link with correct arguments.', function ($data) {
	expect(Helper::getFormTrashActionUrl($data))
		->toBe('id: 0, force: false');
})->with([
    ['0'],
    ['0', 0]
]);;
```

Or without using datasets, you can just put multiple expectations (they look nicer imo).

tests/Helpers/HelperTest.php
```diff
@@ -60,3 +61,123 @@
...
+
+	expect(Helper::logger(['data' => 'Really dangerous error']))->toBeNull();
+});
+
+//---------------------------------------------------------------------------------//
+
+test('minifyString returns expected values', function() {
+	expect(Helper::minifyString('A string that uses'.PHP_EOL.'PHP_EOL'))->toEqual('A string that uses PHP_EOL');
+	expect(Helper::minifyString("A string that uses \r\n\r\n multiple Windows line breaks"))->toEqual("A string that uses \n \n multiple Windows line breaks");
+	expect(Helper::minifyString("A string that\tuses\t\ttabs"))->toEqual("A string that uses tabs");
```
dingo-d, line 107: Could be a dataset, and the expected results can be inside an array (not sure if datasets in pest have a counter like a second argument or something like it).

tests/Validation/ValidatorTest.php
```diff
@@ -0,0 +1,600 @@
...
+				'component' => 'input',
+				'inputName' => 'pattern-is-invalid2',
+				'inputId' => 'pattern-is-invalid2',
+				'inputIsRequired' => true,
+				'inputValidationPattern' => '\\s[0-9]{1,3}',
+			],
+		],  
+	],
+]);
+]);
```
dingo-d, line 600: Is this extra?

dadadavorin (approved):

tests/Validation/ValidatorTest.php
```diff
@@ -660,4 +660,100 @@ public function getOptionValue(string $name): string {
...
 		],  
 	],
+]);
+
+test('validate yields valid results for files', function ($expected, $params, $files, $formId, $formData) {
+	expect($this->validator->validate($params, $files, $formId, $formData))->toBe($expected);
+})->with([
+	[
+		[
+			'file-wrong-type' => 'The file type is not supported. Only .pdf are allowed.',
```
dadadavorin, line 670: Sounds weird a bit :D 
```suggestion
			'file-wrong-type' => 'The file type is not supported. Only .pdf is allowed.',
```

Or maybe "only .pdf's are allowed", "only .pdf files are allowed"

dingo-d:
@mbmjertan can you fix the failing tests, and also adress the phpcs and phpstan issues?

tests/Validation/ValidatorTest.php
mbmjertan: I agree that there's work to be done regarding these labels, but as these are tests, I've went ahead and wrote expectations to be in line with currently implemented labels. I can work on these labels in future commits, but I'd like to focus on increasing test coverage.

I've also considered relying on the appropriate Labels class here (instead of string constants) - not sure what's the better approach here. (cc @dingo-d)

mbmjertan:
@dingo-d Failing tests (which is currently only `validate yields valid results for files` for the `file-wrong-mimetype` param), as well as bugs I discover here, are tracked separately (in https://github.com/infinum/eightshift-forms/issues/223), and I've already fixed some of them. I can fix all issues I find as well, but I'm getting the feeling that this PR might become too much of a chore to review 😄 

As for phpcs and phpstan, they are failing on GitHub Actions because of rate limiting. Can you please take a look at that?

src/Validation/Validator.php
```diff
@@ -297,6 +297,22 @@ private function validateFiles(array $files, array $validationReference, string
 			foreach ($reference as $dataKey => $dataValue) {
 				// Check validation for accept file types.
```
dadadavorin, line 298:
```suggestion
				// Check validation for accepted file types.
```

dadadavorin (approved):

iruzevic (approved):

dingo-d:
@mbmjertan check the latest failures, they weren't rate-limited 🙂

src/Validation/AbstractValidation.php
```diff
@@ -75,9 +77,70 @@ public function isFileMinSizeValid(int $fileSize, int $minFileSize): bool
 	 */
 	public function isFileTypeValid(string $fileName, string $fileTypes): bool
 	{
-		$fileExtension = explode('.', $fileName);
-		$validTypes = explode(',', str_replace(' ', '', str_replace('.', '', $fileTypes)));
+		$validTypes = $this->parseFiletypesString($fileTypes);
+
+		return in_array($this->getFileExtensionFromFilename($fileName), $validTypes, true);
```
dingo-d, line 82: If `$validTypes` is a list (from the documentation it looks like it is), you can do this instead

```suggestion
		return isset($validTypes[$this->getFileExtensionFromFilename($fileName)]);
```

It's more performant than using `in_array` 🙂

src/Validation/AbstractValidation.php
```diff
@@ -75,9 +77,70 @@ public function isFileMinSizeValid(int $fileSize, int $minFileSize): bool
...
+				}
+			}
+		} elseif ($denyIfFileIsNotUploaded) {
+			return false;
+		}
 
-		return in_array(end($fileExtension), $validTypes, true);
+		$fileExtension = $this->getFileExtensionFromFilename($file['name']);
+
+		if (\in_array($fileExtension, \explode('|', $mimeTypes[$fileMimetype] ?? []), true)) {
```
dingo-d, line 115: I'm not a fan of adding extra logic in checks. It makes the code less readable.

I'd `explode` the string first, then check if it's not empty, then do an `isset` check instead of `in_array` 🙂

src/Validation/AbstractValidation.php
```diff
@@ -75,9 +77,70 @@ public function isFileMinSizeValid(int $fileSize, int $minFileSize): bool
...
+	/**
+	 * Parses a comma-separated list of file extensions to return
+	 * a normalized array of allowed extensions.
+	 *
+	 * @param string $fileTypes String of file extensions, e.g. ".pdf, jpg,.gif".
+	 * @return array<string> Array of extensions, e.g. ["pdf", "jpg", "gif"]
+	 */
+	private function parseFiletypesString(string $fileTypes): array
+	{
+		$validTypes = \explode(',', \str_replace(' ', '', \str_replace('.', '', $fileTypes)));
```
dingo-d, line 135: This is kinda like the above. A bit harder to read. Maybe having statements on their own line would be better?

tests/Validation/ValidatorTest.php
dingo-d: Did we discuss this last week? My memory is shoddy 🙈

dingo-d (changes requested):
Left some minor comments 🙂

src/Validation/AbstractValidation.php
mbmjertan: Huh, `$validTypes` is an `array<string>`, e.g. `["pdf", "gif", "jpg"]`:

```
php > $a = ["pdf", "gif", "jpg"];
php > var_dump($a);
array(3) {
  [0]=>
  string(3) "pdf"
  [1]=>
  string(3) "gif"
  [2]=>
  string(3) "jpg"
}
php > var_dump(isset($a["pdf"]));
bool(false)
php > var_dump(in_array("pdf", $a, true));
bool(true)
```

This would work if we used `array_flip` first, but this comes at a performance penalty: `array_flip` has a complexity of O(n), as does `isset` (although it's very fast, it's still linear complexity). 

`in_array` (which is O(n) as well) is slower than `isset`, but faster than `array_flip` + `isset` in benchmarks: https://3v4l.org/c85vN vs. https://3v4l.org/JWD7Q - it's a negligible difference, but one that's in favor of `in_array`. 

I'm all ears for microoptimizations, though 😄

src/Validation/Validator.php
```diff
@@ -295,8 +295,24 @@ private function validateFiles(array $files, array $validationReference, string
...
+							'tmp_name' => $fileValue['tmp_name'][$i],
+						];
+						$individualFiles[] = $file;
+					}
+
+					foreach ($individualFiles as $file) {
+						if (!$this->isMimeTypeValid($file)) {
+							$output[$fileKey] = sprintf($this->labels->getLabel('validationAcceptMime', $formId), $dataValue);
+						}
+					}
```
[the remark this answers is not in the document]
mbmjertan: This is true - `$file` remains in scope of the function, and isn't scoped to the `foreach` loop - but I'm not sure how I feel about your comment. I'm not _against_ it, but I also don't find myself agreeing completely.

The code here is assigning a new value to `$file` upon entering the `foreach` loop (and not consuming it in any way), and we generally take care to reuse variables where it makes sense.  Adding an `unset` would prevent consuming the variable in an unexpected place, but this argument could be used for any single variable we're declaring. It's probably a generally good idea to be more mindful of the ways in which we use and consume variables, but this is an iterator that's reinitialized on its next use. It doesn't seem like an actual concern to me - I don't see how it could cause problems not to unset it if we're _mindful_ ✨  of our variable use and don't consume it for some reason in a different place. Pushing it further, `unset` doesn't free memory, so unsetting it and then declaring a new iterator with the same name could lead to double the memory consumption here. That's probably isn't an actually substantial concern, however.

We also discussed using `reset`, which supposedly comes with performance benefits. I've done some benchmarking on 3v4l, and it seems that neither `unset` or `reset` result in better performance of foreach looping: https://3v4l.org/qfZSk, https://3v4l.org/6B5lg3, https://3v4l.org/EAgOP - test performance does differ between PHP versions (in various ways, which is surprising), but for _most_ PHP versions, it takes `4.0531158447266E-6` to foreach through a 100-element array, no matter whether using `reset`, both `unset` and `reset` or none at all. This is probably partially so because `foreach` itself resets the internal array pointer.

My reasoning regarding this is that I don't see how the benefit of making sure devs don't consume `$file` inbetween the two `foreach` loops outweighs code understandability (with `unset` and `reset` being called), and I don't see any other substantial reasons to use `unset` and none whatsoever to use `reset`. This isn't a hill I'm willing to die upon, though - and if anyone has different feelings regarding this, please let me know.

src/Validation/AbstractValidation.php
dingo-d: Forgot about the flip part 🙈 

This part: 

> has a complexity of O(n), as does isset (although it's very fast, it's still linear complexity).

is not true, because `isset` is O(1) (it just accesses the array to check for existence, while `in_array` is O(n) because it actually iterates over the array).

It's still a microoptimisation part since `in_array` is opcode cached function so it's ok to use it.

iobrado (approved):

dingo-d (approved):

iruzevic (approved):

iruzevic:
I have merged this PR to my and fixed some issues Txn https://github.com/infinum/eightshift-forms/pull/224

PR #240, opened 2022-06-27
adding fixes for ssr

iruzevic:
# Description

- correction fixes for ssr
- adding new event AfterFormEventsClear
- added option to init all scripts.

### DIFF
--- a/src/Blocks/components/form/assets/index.js
+++ b/src/Blocks/components/form/assets/index.js
@@ -4,7 +4,14 @@ import domReady from '@wordpress/dom-ready';
 import { componentJsClass } from './../manifest.json';
 import { FORM_EVENTS, FORM_SELECTORS, FORM_DATA_ATTRIBUTES } from './form';
 
+if (typeof esFormsLocalization === 'undefined') {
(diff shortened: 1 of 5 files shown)


dadadavorin (approved):

dingo-d (approved):

src/Blocks/components/form/assets/index.js
```diff
@@ -4,123 +4,153 @@ import domReady from '@wordpress/dom-ready';
...
+const disableAutoInit = Boolean(esFormsLocalization.formDisableAutoInit) ?? false;
+
 // Load window form no matter what the option is set.
 window['esForms'] = {
 	events: FORM_EVENTS,
 	selectors: FORM_SELECTORS,
 	dataAttributes: FORM_DATA_ATTRIBUTES,
 	formSelector: selector,
+	initAll: () => {
+		initAll(true);
```
volfkarlo, line 23: Sorry to send it now. Might be good to fall back on default but to have an option to pass the option

```
	initAll: (manual = true) => {
		initAll(manual);

iruzevic: if you are triggering this method this will we manual in any way so there is no need for this here

PR #243, opened 2022-07-26
Test coverage improvements

mbmjertan:
# Description

This PR will increase test coverage to about 22%, according to Xdebug.

The PR adds various mocks for WP and PHP functions, which allow for forcing some values (e.g. when using `get_option`) by using environment variables (e.g. `test_force_option_[option name]=[option value]`) and checking that the functions have been called correctly (by providing environment variables such as `test_setcookie_last_call`). 

While Brain Monkey does support [expectations for function calls](https://giuseppe-mazzapica.gitbook.io/brain-monkey/functions-testing-tools/functions-expect), Pest doesn't support tests in isolation, meaning you could have issues when trying to redefine a function two or more times.

It also fixes some bugs encountered when writing tests:
* `getEditorBackLink` could execute `exit` during tests, stopping Pest
* `sortFields` would only work if the number of breakpoints hasn't been changed


dingo-d:
@mbmjertan you mentioned

> While Brain Monkey does support [expectations for function calls](https://giuseppe-mazzapica.gitbook.io/brain-monkey/functions-testing-tools/functions-expect), Pest doesn't support tests in isolation, meaning you could have issues when trying to redefine a function two or more times.

Would this PR be beneficial for this? https://github.com/Brain-WP/BrainMonkey/pull/123?

If so, can you comment on it?

tests/Blocks/BlocksTest.php
```diff
@@ -0,0 +1,128 @@
...
+});
+
+test('Blocks service class registers hooks properly on WP >= 5.8', function () {
+	Actions\expectAdded('init');
+	Filters\expectRemoved('the_content', 'wpautop');
+	Filters\expectAdded('block_categories_all', [$this->blocks, 'getCustomCategory']);
+	Filters\expectAdded('es_blocks_string_to_filter', [$this->blocks, 'getStringToValue']);
+	Filters\expectAdded('es_blocks_options_checkbox_is_checked_filter', [$this->blocks, 'isCheckboxOptionChecked']);
+
+
```
dingo-d, line 35:
```suggestion

```

src/Settings/SettingsHelper.php
```diff
@@ -600,6 +600,12 @@ private function prepareFormViewDetails(array $data): array
 	 */
 	private function sortFields($a, $b): bool
```
dingo-d, line 601: This has to be the worst naming of variables in the history of variable naming 🙈

tests/Editor/EditorTest.php
```diff
@@ -0,0 +1,43 @@
...
+ * Mock before tests.
+ */
+beforeEach(function () {
+	Monkey\setUp();
+	setupMocks();
+	$this->editor = new Editor();
+});
+
+afterEach(function() {
+	Monkey\tearDown();
```
dingo-d, line 21: One thing @iruzevic mentioned to me when I refactored the libs tests is to cleanup the `$this->editor` property.

dingo-d (changes requested):
Left some comments, all in all looking good 👍🏼

tests/Editor/EditorTest.php
mbmjertan: Will do 😄

mbmjertan:
Rebased to `developed` and fixed failing tests, so I'm marking this as ready for review.

dingo-d (approved):

gabriel-glo (approved):

PR #244, opened 2022-08-01
Hubspot integration fixes

iruzevic:
# Description

## Fixed
- Preselected values for custom select

## Added
- new fieldHidden attributes for hiding fields from dom.
- hidden field for Hubspot integration.
- preselected value field for Hubspot integration.
- enabled field for Hubspot integration.


volfkarlo (approved):
This looks promising! 😄 👏

PR #245, opened 2022-08-02
Custom form actions

gabriel-glo:
Added a new option field to set a custom form action to which the form will send a request.
Updated form handlers to retrieve an process form action attribute
Updated form render to set the form type based on the form action

Added a new custom route for handling forms with custom actions.
Route takes all parameters and POSTs them to the provided action URL. Response from the action URL is processed and generic messages are sent.


piqusy (commented):
Looks good to me. Can we please have PHPCS errors resolved?

iruzevic (changes requested):
looking good please fix linting issues

PR #246, opened 2022-08-03
Feature/localstorage

iruzevic:
# Description

### Added
- new tracking class for storing url tags and detecting tags from GET param.
- New option to store url tracking tags to local storage for later usage.
- New filters for tracking.

### DIFF
--- a/src/Hooks/FiltersTracking.md
+++ b/src/Hooks/FiltersTracking.md
@@ -1,0 +1,75 @@
+# Filters Tracking
+This document will provide you with the code examples for forms filters used in tracking.
+
+## Add additional tags
+This filter provides you with the ability to add/remove/edit url tags sent to the backend when processing the form request.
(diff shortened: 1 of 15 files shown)


src/Hooks/FiltersTracking.md
```diff
@@ -73,3 +73,25 @@ public function getIntegrationHubspotLocalStorageMap(array $params, array $stora
...
+```php
+// Get tracking expiration days.
+\add_filter('es_forms_tracking_expiration', [$this, 'getTrackingExpiration']);
+
+/**
+ * Get tracking expiration days.
+ *
+ * @return string
+ */
+public function getTrackingExpiration(): string
```
dadadavorin, line 93: why not add a parameter here so when calling a filter it can be any value and not fixed?

iruzevic: no sure what you are getting at, can you please provide little more context

PR #247, opened 2022-08-08
adding hubspot upload allow type limit

iruzevic:
# Description

### Added
- New global and local settings for limiting HubSpot file upload mime type.

### DIFF
--- a/src/Blocks/components/form/assets/form.js
+++ b/src/Blocks/components/form/assets/form.js
@@ -1071,15 +1071,8 @@ export class Form {
 
 		// Find url params.
 		const searchParams = new URLSearchParams(window.location.search);
-		for (const [key, value] of searchParams.entries()) {
-			// Bailout if not allowed or empty
-			if (!allowedTags.includes(key) || value === '') {
-				continue;
-			}
-
-			// Add valid tag.
-			newStorage[key] = value;
-		}
+		const newStorage = searchParams.entries()
+			.filter(([key, value]) => allowedTags.includes(key) && value !== '')
 
 		// Bailout if nothing is set from allowed tags or everything is empty.
 		if (Object.keys(newStorage).length === 0) {
--- a/src/Blocks/components/form/assets/form.js
+++ b/src/Blocks/components/form/assets/form.js
@@ -77,7 +77,7 @@ export class Form {
 		this.hideLoadingStateTimeout = options.hideLoadingStateTimeout ?? 600;
 		this.fileCustomRemoveLabel = options.fileCustomRemoveLabel ?? '';
 		this.captcha = options.captcha ?? '';
-		this.storage = options.storage ?? '';
+		this.storageConfig = options.storageConfig ?? '';
 
 		// Internal state.
 		this.files = {};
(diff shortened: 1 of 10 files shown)


src/Validation/SettingsCaptcha.php
```diff
@@ -212,7 +212,7 @@ public function getSettingsGlobalData(): array
 						'component' => 'input',
 						'inputId' => $this->getSettingsName(self::SETTINGS_CAPTCHA_SCORE_KEY),
 						'inputFieldLabel' => \__('"Spam unlikely" threshold', 'eightshift-forms'),
-						'inputFieldHelp' => \__('This number determines the level above which a submission is <strong>not</strong> considered spam. <br /> <br /> The value should be between 0.0 and 1.0 (default is 0.5).', 'eightshift-forms'),
+						'inputFieldHelp' => \__('This number determines the level above which a submission is <strong>not</strong> considered spam. In general normal users will always get score of 0.8-0.9. The value should be between 0.1 and 1.0 (default is 0.5).', 'eightshift-forms'),
```
mbmjertan, line 215:
```suggestion
						'inputFieldHelp' => \__('This number determines the level above which a submission is <strong>not</strong> considered spam. In general, normal users will get a score of around 0.8-0.9. The value should be between 0.1 and 1.0 (default is 0.5).', 'eightshift-forms'),
```

mbmjertan (approved):
Looks good to me!

src/Blocks/components/form/assets/form.js
```diff
@@ -1047,26 +1047,47 @@ export class Form {
...
+		const searchParams = new URLSearchParams(window.location.search);
+		for (const [key, value] of searchParams.entries()) {
+			// Bailout if not allowed or empty
+			if (!allowedTags.includes(key) || value === '') {
+				continue;
+			}
+
+			// Add valid tag.
+			newStorage[key] = value;
+		}
```
volfkarlo, line 1082:
```suggestion
		const newStorage = searchParams.entries()
			.filter(([key, value]) => allowedTags.includes(key) && value !== '')
```

src/Blocks/components/form/assets/form.js
```diff
@@ -1047,26 +1047,47 @@ export class Form {
...
+		// Bailout if nothing is set from allowed tags or everything is empty.
+		if (Object.keys(newStorage).length === 0) {
+			return;
+		}
+
+		// Add current timestamp to new storage.
+		newStorage.timestamp = Date.now();
 
 		// Store in a new variable for later usage.
 		const newStorageFinal = {...newStorage};
```
volfkarlo, line 1093:
```suggestion
		// Store in a new variable for later usage.
		const newStorageFinal = {...newStorage};
		newStorage.timestamp = Date.now();
```
Wouldn't this avoid the delete process on `newStorageFinal`?

volfkarlo (changes requested):
Mini comments, great job! 👏

src/Blocks/components/form/assets/form.js
iruzevic: newStorage is used to compare the localStorages and the newStorageFinal is used to have an reference to the original one. This is ok

volfkarlo: Yes, but if you change the order, you don't have to delete it.
If you are ok with deleting, all good, this was just a suggestion

volfkarlo (approved):
👏

PR #307, opened 2023-06-06
Adding fix for sync diff if there is an error with the integration

iruzevic:
# Description

- I have implemented a solution to prevent the deletion of forms in cases where the output is empty during differential synchronization. This ensures that forms remain accessible even if the integrations fail.

### DIFF
--- a/src/Blocks/components/utils/index.js
+++ b/src/Blocks/components/utils/index.js
@@ -16,6 +16,18 @@ import { AnimatedContentVisibility, camelize, classnames, IconLabel, icons, STOR
  *
  * @returns {boolean}
  */
+export const isDeveloperMode = () => {
+	return Boolean(esFormsLocalization?.isDeveloperMode);
+};
+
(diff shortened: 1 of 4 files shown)


src/Blocks/components/utils/index.js
```diff
@@ -55,23 +67,25 @@ export const updateIntegrationBlocks = (clientId, postId, type, itemId, innerId
  */
 export const syncIntegrationBlocks = (clientId, postId) => {
 	return apiFetch({ path: `${esFormsLocalization.restPrefixProject}${esFormsLocalization.restRoutes.integrationsEditorSync}/?id=${postId}` }).then((response) => {
-		resetInnerBlocks(clientId);
+		if (isDeveloperMode()) {
+			console.log(response);
```
iruzevic, line 71: this is intentional

piqusy (approved):
Looks good to me 👍🏼

PR #359, opened 2024-01-25
adding new debug options

iruzevic:
# Description

## Added
- new debug options for form field
- new debug options for steps

## Removed
- utils icons are now used from utils lib


iobrado (approved):

src/Blocks/components/form/assets/step.js
```diff
@@ -327,6 +327,16 @@ export class Steps {
...
 	}
 
+	toggleDebugPreview() {
+		const debug = document.querySelectorAll(this.state.getStateSelector('stepDebugPreview', true));
+
+		if (debug.length) {
+			[...debug].forEach((item) => {
+				item.addEventListener('click', this.onToggleDebugPreview);
+			});
+		}
```
goranalkovic-infinum, line 337:
```suggestion
		const debug = document.querySelectorAll(this.state.getStateSelector('stepDebugPreview', true));

		debug.forEach((item) => item?.addEventListener('click', this.onToggleDebugPreview));
```
Since you're using `querySelectorAll`, you can be sure that you'll always get a `NodeList`, which is iterable just with `forEach` directly 😉 

image

src/Troubleshooting/SettingsDebug.php
```diff
@@ -167,6 +167,7 @@ public function getSettingsGlobalData(): array
 									<ul>
 										<li>Every listing will have ID prepended to the label.</li>
 										<li>Integration api response will have a `debug` key with all the details. You can check it out using inspector network tab.</li>
```
goranalkovic-infinum, line 169:
```suggestion
										<li>Integration API response will have a `debug` key with all the details. You can check it out using inspector network tab.</li>
```

src/Troubleshooting/SettingsDebug.php
```diff
@@ -167,6 +167,7 @@ public function getSettingsGlobalData(): array
...
+										<li>On frontend form when hovering over the field you will see debug tooltip with some helpful information.</li>
```
goranalkovic-infinum, line 170:
```suggestion
										<li>On the frontend, when hovering a over a form field a debug tooltip will be shown with helpful information regarding that field.</li>
```

goranalkovic-infinum (changes requested):

piqusy (dismissed):
Nice work 👏🏼

kancijan (approved):

PR #360, opened 2024-01-31
Feature/rtl support

piqusy:
# Description

Adding RTL (Right-To-Left) support by updating CSS to use logical properties.

Docs:
[CSS logical properties and values](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_logical_properties_and_values)


iruzevic (approved):
looking good

PR #361, opened 2024-01-31
Feature/rtl support - update border-radius

piqusy:
# Description

Update border-radius properties


goranalkovic-infinum (approved):
as Dragana M. would say: _because I trusted you blindly_
(too many of the same declarations to read file-by-file without eyes and brain hurting)

iruzevic (approved):

PR #367, opened 2024-02-07
adding new filters for external events

iruzevic:
# Description

- new filters
- moved function to helper utils

### DIFF
--- a/src/Rest/Routes/Integrations/Moments/FormSubmitMomentsRoute.php
+++ b/src/Rest/Routes/Integrations/Moments/FormSubmitMomentsRoute.php
@@ -141,7 +141,7 @@ class FormSubmitMomentsRoute extends AbstractFormSubmit
 		$eventName = UtilsSettingsHelper::getSettingValue(SettingsMoments::SETTINGS_MOMENTS_EVENTS_EVENT_NAME_KEY, $formId);
 		$map = UtilsSettingsHelper::getSettingValueGroup(SettingsMoments::SETTINGS_MOMENTS_EVENTS_MAP_KEY, $formId);
 
-		if (!$emailKey || !$eventName || !$map) {
+		if (!$emailKey || !$eventName) {
 			return;
 		}
 
--- a/composer.json
+++ b/composer.json
@@ -40,7 +40,7 @@
 	"require": {
 		"php": "^7.4 || >=8.0",
 		"erusev/parsedown": "^1.7.4",
-		"infinum/eightshift-forms-utils": "^1.1.4"
+		"infinum/eightshift-forms-utils": "^1.1.5"
 	},
 	"suggest": {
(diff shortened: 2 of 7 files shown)


kancijan (approved):

goranalkovic-infinum (approved):

src/Rest/Routes/Integrations/Moments/FormSubmitMomentsRoute.php
```diff
@@ -141,7 +141,7 @@ private function sendEvent(array $formDetails): void
 		$eventName = UtilsSettingsHelper::getSettingValue(SettingsMoments::SETTINGS_MOMENTS_EVENTS_EVENT_NAME_KEY, $formId);
 		$map = UtilsSettingsHelper::getSettingValueGroup(SettingsMoments::SETTINGS_MOMENTS_EVENTS_MAP_KEY, $formId);
 
-		if (!$emailKey || !$eventName || !$map) {
+		if (!$emailKey || !$eventName) {
```
piqusy, line 144:
```suggestion
		if (!($emailKey && $eventName)) {
```

piqusy (approved):
Nice work 👏🏼 

Just a minor suggestion

src/Rest/Routes/Integrations/Moments/FormSubmitMomentsRoute.php
iruzevic: this is not correct:

```
No, `!$emailKey || !$eventName` is not the same as `!($emailKey && $eventName)`.

Let's break down each expression:

1. `!$emailKey || !$eventName`: This expression evaluates to `true` if either `$emailKey` is false (or null or empty) OR `$eventName` is false (or null or empty). In other words, it will be true if either of the variables is falsy.

2. `!($emailKey && $eventName)`: This expression first evaluates the logical AND operator `&&` between `$emailKey` and `$eventName`. If both `$emailKey` and `$eventName` are true (not falsy), then the expression inside the parentheses evaluates to `true`. The `!` operator then negates this result, so the entire expression becomes `false`. However, if either `$emailKey` or `$eventName` (or both) is false, the expression inside the parentheses evaluates to `false`, and the `!` operator negates it to become `true`.

So, these expressions are not equivalent. The first expression checks if either variable is falsy, while the second expression checks if both variables are truthy.

```

piqusy: The same guy said to me:

```
Given the original expression !$emailKey || !$eventName, the De Morgan's Law transformation should involve switching between conjunction (AND) and disjunction (OR), along with negating each individual term. Here's the correct transformation:

Original expression: !$emailKey || !$eventName

De Morgan's Law transformation: !( $emailKey && $eventName )

Therefore, the correct equivalent expression using De Morgan's Law is !( $emailKey && $eventName ). Thank you for your patience and for bringing that to my attention.
```

😄

PR #400, opened 2024-04-15
updates on reCaptcha

iruzevic:
### Fixed
- reCaptcha will now output the correct msg on the frontend in any case.

### Added
- reCaptcha will now send fallback email if there is an issue with the validation.
- Security feature now supports option to ignore IPs.

### DIFF
--- a/src/Labels/Labels.php
+++ b/src/Labels/Labels.php
@@ -401,16 +401,10 @@ class Labels implements LabelsInterface
 	private function getCaptchaLabels(): array
 	{
 		return [
-			'captchaMissingInputSecret' => \__('The Captcha "secret" parameter is missing.', 'eightshift-forms'),
-			'captchaInvalidInputSecret' => \__('The Captcha "secret" parameter is invalid or malformed.', 'eightshift-forms'),
-			'captchaInvalidInputResponse' => \__('The Captcha "response" parameter is invalid or malformed.', 'eightshift-forms'),
-			'captchaMissingInputResponse' => \__('The Captcha "response" parameter is missing.', 'eightshift-forms'),
-			'captchaBadRequest' => \__('The Captcha "request" is invalid or malformed.', 'eightshift-forms'),
-			'captchaTimeoutOrDuplicate' => \__('The Captcha response is no longer valid: either is too old or has been used previously.', 'eightshift-forms'),
-			'captchaWrongAction' => \__('The Captcha response "action" is not valid.', 'eightshift-forms'),
-			'captchaIncorrectCaptchaSol' => \__('The Captcha keys are not valid. Please check your site and secret key configuration.', 'eightshift-forms'),
-			'captchaScoreSpam' => \__('The automated system detected this request as a potential spam request. Please try again.', 'eightshift-forms'),
-			'captchaBrowserError' => \__('The Captcha check encountered an error. Please try again.', 'eightshift-forms'),
+			'captchaBadRequest' => \__('Spam prevention system encountered an error. Captcha "request" is invalid or malformed.', 'eightshift-forms'),
+			'captchaWrongAction' => \__('Spam prevention system encountered an error. Captcha response "action" is not valid.', 'eightshift-forms'),
+			'captchaScoreSpam' => \__('Our system detected this request as a potential spam request. Please try again.', 'eightshift-forms'),
+			'captchaError' => \__('Spam prevention system encountered an error. Please try again.', 'eightshift-forms'),
 		];
 	}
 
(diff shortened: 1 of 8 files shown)


dadadavorin (approved):
👏

dadadavorin (dismissed):

src/Security/Security.php
```diff
@@ -47,9 +48,19 @@ public function isRequestValid(): bool
...
-		$ip = $this->getIpAddress(true);
+		$ip = $this->getIpAddress();
 		$time = \time();
 
+		// Bailout if the IP is in the ignore list.
+		$ignoreIps = Components::flattenArray(UtilsSettingsHelper::getOptionValueGroup(SettingsSecurity::SETTINGS_SECURITY_IP_IGNORE_KEY));
+
+		if (\array_key_exists($ip, \array_flip($ignoreIps))) {
+			return true;
+		}
```
dingo-d, line 59: Could this be written with `isset`?

```suggestion
		if (isset(\array_flip($ignoreIps)[$ip])) {
			return true;
		}
```

Not sure what the `$ignoreIps` array looks like, but unless the array is like:

```php
[
  '123.456.789.0' => null,
   ...
]
```

You can use `isset`. I kinda reckon that since you are flipping the array you'll never have null inside (otherwise you'd loose the key that is `null`).

src/Validation/Validator.php
```diff
@@ -260,13 +260,13 @@ public function validateParams(array $formDetails, bool $strictValidation = true
...
 					case 'min':
-						if ((string) $dataValue > (string) $inputValue) {
+						if ((string) $dataValue > (string) $inputValue && !empty($inputValue)) {
 							$output[$paramKey] = \sprintf($this->getValidationLabel('validationMin', $formId), $dataValue);
 						}
 						break;
 					// Check validation for min number value.
 					case 'max':
-						if ((string) $dataValue < (string) $inputValue) {
+						if ((string) $dataValue < (string) $inputValue && !empty($inputValue)) {
```
dingo-d, line 269: Would it make more sense to bail out earlier if the `$inputValue` is empty?

Also, why are you comparing strings? I mean they will be numerically compared, but still I think you said that you're comparing numbers.

dingo-d (changes requested):
Left a suggestion and a question.

src/Validation/Validator.php
```diff
@@ -260,13 +260,13 @@ public function validateParams(array $formDetails, bool $strictValidation = true
 						break;
 					// Check validation for min number value.
 					case 'min':
-						if ((string) $dataValue > (string) $inputValue) {
+						if ((string) $dataValue > (string) $inputValue && !empty($inputValue)) {
```
mbmjertan, line 263: assume min number value is `42` and submission value is `112` - I think this would be false?

src/Security/Security.php
iruzevic: sure looks good

src/Validation/Validator.php
iruzevic: this is tested and it work

src/Validation/Validator.php
iruzevic: This can be a minus number, float or empty so it was easier to compare strings. 

for empty check this will validate only if value is populated so we don't validate empty fields that are not required.

dingo-d (dismissed):

src/Labels/Labels.php
```diff
@@ -401,16 +401,10 @@ private function getActiveCampaignLabels(): array
...
-			'captchaBadRequest' => \__('The Captcha "request" is invalid or malformed.', 'eightshift-forms'),
-			'captchaTimeoutOrDuplicate' => \__('The Captcha response is no longer valid: either is too old or has been used previously.', 'eightshift-forms'),
-			'captchaWrongAction' => \__('The Captcha response "action" is not valid.', 'eightshift-forms'),
-			'captchaIncorrectCaptchaSol' => \__('The Captcha keys are not valid. Please check your site and secret key configuration.', 'eightshift-forms'),
-			'captchaScoreSpam' => \__('The automated system detected this request as a potential spam request. Please try again.', 'eightshift-forms'),
-			'captchaBrowserError' => \__('The Captcha check encountered an error. Please try again.', 'eightshift-forms'),
+			'captchaBadRequest' => \__('Spam prevention system encountered an error. Captcha "request" is invalid or malformed.', 'eightshift-forms'),
+			'captchaWrongAction' => \__('Spam prevention system encountered an error. Captcha response "action" is not valid.', 'eightshift-forms'),
+			'captchaScoreSpam' => \__('Our system detected this request as a potential spam request. Please try again.', 'eightshift-forms'),
```
goranalkovic-infinum, line 406:
```suggestion
			'captchaScoreSpam' => \__('The request was marked as a potential spam request. Please try again.', 'eightshift-forms'),
```

PR #426, opened 2024-06-04
Feature/number input

iruzevic:
### Added
- Input number validation for allowed keys.

### DIFF
--- a/src/Blocks/components/form/assets/form.js
+++ b/src/Blocks/components/form/assets/form.js
@@ -1018,6 +1018,8 @@ export class Form {
 		input.addEventListener('keydown', this.onFocusEvent);
 		input.addEventListener('focus', this.onFocusEvent);
 		input.addEventListener('blur', this.onBlurEvent);
+		input.addEventListener('keydown', this.onKeyDownEvent);
+
 
 		// If field range, set current value to DOM.
 		if (this.state.getStateElementTypeCustom(name, formId) === 'range') {
@@ -1477,6 +1479,7 @@ export class Form {
 				input?.removeEventListener('focus', this.onFocusEvent);
 				input?.removeEventListener('blur', this.onBlurEvent);
 				input?.removeEventListener('input', this.onInputEvent);
+				input?.removeEventListener('keydown', this.onKeyDownEvent);
 			});
 
 			// Date.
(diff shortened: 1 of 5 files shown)


src/Blocks/components/form/assets/form.js
```diff
@@ -1658,6 +1661,46 @@ export class Form {
...
+	 */
+	onKeyDownEvent = (event) => {
+		const formId = this.state.getFormIdByElement(event.target);
+		const field = this.state.getFormFieldElementByChild(event.target);
+		const name = field.getAttribute(this.state.getStateAttribute('fieldName'));
+
+		if (this.state.getStateElementTypeCustom(name, formId) === 'number') {
+			const allowedKeys = [
+				'Backspace',
+				'Enter',
```
goranalkovic-infinum, line 1679: Can't you just filter things when setting state/submitting the form?
This seems a bit hacky and could maybe be bypassed with assistive technologies etc

goranalkovic-infinum (changes requested):

src/Blocks/components/form/assets/form.js
iruzevic: the problem is that I don't have anything to filter as if you get the value with text the value will be empty 👎 and I can't filter anything :(((

iobrado (approved):

dadadavorin (approved):

goranalkovic-infinum (approved):

src/Blocks/components/form/assets/form.js
```diff
@@ -1658,6 +1661,46 @@ export class Form {
...
+	 *
+	 * @returns {void}
+	 */
+	onKeyDownEvent = (event) => {
+		const formId = this.state.getFormIdByElement(event.target);
+		const field = this.state.getFormFieldElementByChild(event.target);
+		const name = field.getAttribute(this.state.getStateAttribute('fieldName'));
+
+		if (this.state.getStateElementTypeCustom(name, formId) === 'number') {
+			const allowedKeys = [
```
piqusy, line 1677: How would this hold against the [Chinese numerals](https://en.wikipedia.org/wiki/Chinese_numerals) ?

iruzevic: I have no idea really and no way to test this :(

piqusy: Seems like Chrome (form > input field) does not support these.
From what I have found it seems that Arabic numbers are also a common method for numeric inputs in Chinese language

I'd say this is more than acceptable 😄

piqusy (approved):

PR #443, opened 2024-08-27
adding tailwind filter

iruzevic:
# Description

### Added
- Filter to load Tailwind classes on all elements.

### DIFF
--- a/src/Hooks/FiltersOuputMock.php
+++ b/src/Hooks/FiltersOuputMock.php
@@ -284,6 +284,34 @@ final class FiltersOuputMock
 		];
 	}
 
+	/**
+	 * Return Tailwind selectors data filter output.
+	 *
+	 * @param string $selector Selector to get data for.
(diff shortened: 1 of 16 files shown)


dadadavorin (approved):
Looks good to me 👌

vedranpetrac (approved):

src/Hooks/FiltersOuputMock.php
piqusy, line 1: I'm trying to figure out why this file was named like this but to no avail.
Can you please provide some insight on that?

piqusy (approved):

src/Hooks/FiltersOuputMock.php
iruzevic: As we have a bunch of filters all over the place,  the idea of this file is to group filters in one place and provide output data like global settings, local settings and filter settings.

piqusy: OK, but why `...Mock` 🤔

PR #507, opened 2025-04-25
Release 7

iruzevic:


mbmjertan:
>  deprecate forms-utils

What are the implications of this on plugins relying on forms-utils?

iruzevic:
> > deprecate forms-utils
> 
> What are the implications of this on plugins relying on forms-utils?

Forms will no longer update the utils library or rely on it. You can copy all the code you need into your plugin, and everything should work fine. The utils library will remain active until you finish the switch.

PR #610, opened 2026-04-03
Sanitize NationBuilder field values

ghost:
## Summary
- Sanitized NationBuilder field values with `wp_strip_all_tags` to prevent HTML injection into NationBuilder fields
- Bumped version to 9.1.1 and updated changelog

## Test plan
- [ ] Submit a form with HTML tags in input fields mapped to NationBuilder
- [ ] Verify the tags are stripped before being sent to NationBuilder API
- [ ] Verify normal text submissions still work correctly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 This projects adheres to [Semantic Versioning](https://semver.org/) and [Keep a CHANGELOG](https://keepachangelog.com/).
 
+## [9.1.1]
+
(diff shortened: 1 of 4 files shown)


goranalkovic-infinum (approved):

piqusy (changes requested):
Is NationBuilder the only one with this `hole`?

iruzevic (approved):

kancijan:
@piqusy not sure, but the client requested this one

piqusy (approved):

PR #613, opened 2026-04-16
Friendly Captcha

iobrado:
# Description

## Friendly Captcha v2 Integration

Adds Friendly Captcha v2 as a privacy-focused alternative to Google reCAPTCHA for spam prevention.

### Why

Google reCAPTCHA relies on tracking and behavioral profiling, which raises GDPR/privacy concerns for EU-based clients. Friendly Captcha uses proof-of-work puzzles instead — no cookies, no tracking, no user interaction required. This gives clients a compliant option without sacrificing spam protection.

### What's included

- Unified settings page under Forms → Settings → Spam prevention (Advanced section) with a provider dropdown to switch between Google reCAPTCHA and Friendly Captcha — only one can be active at a time
- Friendly Captcha settings include site key and API key fields with environment variable support
- Zero-click experience — the widget runs invisibly in the background using [startMode: auto](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), similar to reCAPTCHA v3
- Full integration with all form submission routes — works with Mailer, HubSpot, Mailchimp, Greenhouse, and all other integrations out of the box
- Server-side verification via Friendly Captcha's /api/v2/captcha/siteverify endpoint, dispatched through a new CaptchaDispatcher that delegates to the active provider at runtime
- Single CaptchaInterface contract — FriendlyCaptchaInterface was removed; both providers implement the same interface
- Removed AbstractCaptchaProvider — both Captcha and FriendlyCaptcha are fully self-contained
- Activity log support with dedicated fallback flags for troubleshooting
- Developer mode — skip-captcha bypasses Friendly Captcha the same way it does reCAPTCHA

### No breaking changes

All existing reCAPTCHA functionality is untouched.

# Screenshots / Videos

Screenshot 2026-04-21 at 10 57 24

# Linked documentation PR

https://developer.friendlycaptcha.com/docs/v2/getting-started/


src/Blocks/components/form/assets/form.js
```diff
@@ -1157,6 +1179,22 @@ export class Form {
 		], this.FORM_DATA);
 	}
 
+	/**
+	 * Set form data object for all forms - friendly captcha.
```
piqusy:
```suggestion
	 * Set form data object for all forms - Friendly Captcha.
```

src/Blocks/components/form/assets/form.js
```diff
@@ -1818,6 +1856,8 @@ export class Form {
 
 					if (this.state.getStateCaptchaIsUsed()) {
```
piqusy, line 1837: Would it be a major breaking change to update this method name to be less generic now that we have more than one captcha?

src/FriendlyCaptcha/SettingsFriendlyCaptcha.php
```diff
@@ -0,0 +1,234 @@
...
+		$apiKey = (bool) SettingsHelpers::getOptionWithConstant(Variables::getFriendlyCaptchaApiKey(), self::SETTINGS_FRIENDLY_CAPTCHA_API_KEY);
+
+		if (!$isUsed || !$siteKey || !$apiKey) {
+			return false;
+		}
+
+		// Mutual exclusivity: if Google reCAPTCHA is also valid, Friendly Captcha is not valid.
+		if (\apply_filters(SettingsCaptcha::FILTER_SETTINGS_GLOBAL_IS_VALID_NAME, false)) {
+			return false;
+		}
```
piqusy: This one-sided check seems fragile

src/FriendlyCaptcha/SettingsFriendlyCaptcha.php
```diff
@@ -0,0 +1,234 @@
...
+				SettingsOutputHelpers::getIntro(self::SETTINGS_TYPE_KEY),
+				[
+					'component' => 'highlighted-content',
+					'highlightedContentTitle' => \__('Google reCAPTCHA is active', 'eightshift-forms'),
+					// phpcs:ignore WordPress.WP.I18n.NoHtmlWrappedStrings
+					'highlightedContentSubtitle' => \__('Please disable Google reCAPTCHA before configuring Friendly Captcha.<br />Only one spam prevention provider can be active at a time.', 'eightshift-forms'),
+					'highlightedContentIcon' => 'warning',
+				],
+			];
+		}
```
piqusy: Same concern - the  check only exists on FC side

piqusy (commented):
Nice work 👏🏼 

Left a few suggestions / questions

iruzevic (changes requested):
As we discussed on the 1-on-1 call, this feature will need a somewhat different architectural approach. When you get a chance, could you take a look at implementing it along those lines? Feel free to ping me if you run into any questions or issues along the way.

src/FriendlyCaptcha/SettingsFriendlyCaptcha.php
iobrado: This was reworked, not a concern anymore 🙂

src/Blocks/components/form/assets/form.js
iobrado: It would be a bit easier to read, as now it's kinda confusing, but then all Google reCaptcha related methods that use generic Captcha name should be renamed.

iobrado:
@iruzevic I implemented the changes we discussed. Please let me know if I missed anything or if anything else needs further changes.

Thanks!

src/Blocks/components/form/assets/form.js
piqusy: A diplomatic reply but with no firm resolution 😄

piqusy (changes requested):
One outstanding question 😄

src/Blocks/components/form/assets/form.js
iobrado: That's the "I'm scared of breaking something" reply 😂

piqusy: No worries, just bump the major 😉

src/Blocks/components/form/assets/form.js
```diff
@@ -1818,6 +1861,8 @@ export class Form {
...
 						this.runFormCaptcha(formId, filterFinal);
+					} else if (this.state.getStateFriendlyCaptchaIsUsed()) {
```
iruzevic: We don't need this check here — `getStateCaptchaIsUsed` should be the only check we need, just like before, regardless of which captcha is used. That way you have a single check, and then inside `runFormCaptcha` you can check which type is being used. The same in other places not just here

src/Blocks/components/form/assets/friendly-captcha.js
```diff
@@ -0,0 +1,110 @@
...
+		this.initWidget();
+	}
+
+	/**
+	 * Initialize Friendly Captcha widget.
+	 *
+	 * @returns {void}
+	 */
+	initWidget() {
+		const siteKey = this.state.getStateFriendlyCaptchaSiteKey();
```
iruzevic: This should probably go after the `typeof` check, since `frcaptcha` is missing — the rest becomes less relevant.

src/Blocks/components/form/assets/friendly-captcha.js
```diff
@@ -0,0 +1,110 @@
...
+		if (typeof frcaptcha === 'undefined') {
+			return;
+		}
+
+		// Create a hidden container for the widget.
+		const container = document.createElement('div');
+		container.style.display = 'none';
+		document.body.appendChild(container);
+
+		this.widget = frcaptcha.createWidget({
```
iruzevic, line 71: You should initialize `this.widget` to a null or empty object in the constructor, since it's referenced elsewhere in the class.

src/Blocks/components/form/assets/friendly-captcha.js
```diff
@@ -0,0 +1,110 @@
...
+			apiEndpoint: this.state.getStateFriendlyCaptchaEndpoint(),
+		});
+	}
+
+	/**
+	 * Get the current response token from the widget.
+	 *
+	 * @returns {string} The solution token.
+	 */
+	getResponse() {
```
iruzevic: where is this used?

src/Blocks/components/form/assets/friendly-captcha.js
```diff
@@ -0,0 +1,110 @@
...
+		return this.widget?.getResponse() ?? '';
+	}
+
+	/**
+	 * Reset the widget so it generates a fresh token.
+	 *
+	 * @returns {void}
+	 */
+	reset() {
```
iruzevic: where is this used?

src/Blocks/components/form/assets/index.js
```diff
@@ -24,6 +24,13 @@ domReady(() => {
 		});
 	}
 
+	// Load Friendly Captcha if used.
+	if (state.getStateFriendlyCaptchaIsUsed()) {
+		import('./friendly-captcha').then(({ FriendlyCaptcha }) => {
```
iruzevic: With Google reCAPTCHA the early init loads it on every page since it can be used for more than just forms. Is this the same approach here? If so, I'm not seeing a toggle for this in the admin. If it is not this captcha should only be loaded on forms pages

Image

src/Blocks/components/select/select-admin.scss
```diff
@@ -43,5 +43,6 @@ div.es-select {
 		--es-input-radius: var(--global-es-radius-8) var(--global-es-radius-8) 0 0;
 
 		box-shadow: var(--global-esf-box-shadow-l);
+		z-index: 2;
```
iruzevic: real specific, what was the issue here?

src/Blocks/components/form/assets/form.js
```diff
@@ -1157,6 +1184,22 @@ export class Form {
...
+	 * Set form data object for all forms - Friendly Captcha.
+	 *
+	 * @param {object} data Friendly Captcha data.
+	 *
+	 * @returns {void}
+	 */
+	setFormDataFriendlyCaptcha(data) {
+		this.utils.buildFormDataItems([
+			{
+				name: this.state.getStateParam('friendlyCaptcha'),
```
iruzevic: could we use the same parameter name for all captchas? I noticed you're mapping it back to the original name in the PHP anyway, so it seems like it might be possible to unify that.

src/Blocks/manifest.json
```diff
@@ -770,6 +771,7 @@
...
+		"friendlyCaptcha": "<svg xmlns='http://www.w3.org/2000/svg' width='20' height='20' viewBox='0 0 20 20' fill='none'><path d='M10 1 3 4v5c0 5.25 3.5 8.75 7 10 3.5-1.25 7-4.75 7-10V4l-7-3Z' stroke='currentColor' stroke-linejoin='round' fill='none'/><path d='M10 1 3 4v5c0 5.25 3.5 8.75 7 10 3.5-1.25 7-4.75 7-10V4l-7-3Z' fill='currentColor' fill-opacity='0.12'/><path d='m7 10 2 2 4-4' stroke='currentColor' stroke-linecap='round' stroke-linejoin='round' fill='none'/></svg>",
```
iruzevic: you are probably not using this as we have only one icon for captcha

src/Captcha/CaptchaDispatcher.php
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace EightshiftForms\Captcha;
+
+/**
+ * CaptchaDispatcher class.
+ */
+class CaptchaDispatcher implements CaptchaInterface
```
iruzevic: not the best approach:
1. You should have left the original Captcha.php as the dispatcher. In this way, you don't have to refactor anything later in the Routes
2. The original Captcha.php should be named CaptchaRecaptcha or something like you did with FriendlyCaptcha

src/Captcha/FriendlyCaptcha.php
```diff
@@ -0,0 +1,179 @@
...
+	 * @param string $action Action to check (unused).
+	 * @param boolean $isEnterprise Type of captcha (unused).
+	 * @param array<string, mixed> $formDetails Form details.
+	 *
+	 * @return array<mixed>
+	 */
+	public function check(string $token, string $action, bool $isEnterprise, array $formDetails = []): array
+	{
+		if (!\apply_filters(SettingsFriendlyCaptcha::FILTER_SETTINGS_GLOBAL_IS_VALID_NAME, false)) {
+			return $this->buildSuccess(SettingsFallback::SETTINGS_FALLBACK_FLAG_FRIENDLY_CAPTCHA_FEATURE_DISABLED, []);
```
iruzevic: Please follow the same logic as we have in all other classes. No need to extract everything to an additional function.

src/Captcha/FriendlyCaptcha.php
```diff
@@ -0,0 +1,179 @@
...
+		}
+
+		$debug = [
+			'token' => $token,
+			'formDetails' => $formDetails,
+		];
+
+		if (!$token) {
+			$this->throwError(
```
iruzevic: the same as earlier, these are all helpers that make this class harder to read please simplify it as it is in other classes

src/Captcha/FriendlyCaptcha.php
```diff
@@ -0,0 +1,179 @@
...
+	 *
+	 * @return array<mixed>|WP_Error
+	 */
+	private function remoteCall(string $token): array|WP_Error
+	{
+		$siteKey = SettingsHelpers::getOptionWithConstant(Variables::getFriendlyCaptchaSiteKey(), SettingsFriendlyCaptcha::SETTINGS_FRIENDLY_CAPTCHA_SITE_KEY);
+		$apiKey = SettingsHelpers::getOptionWithConstant(Variables::getFriendlyCaptchaApiKey(), SettingsFriendlyCaptcha::SETTINGS_FRIENDLY_CAPTCHA_API_KEY);
+
+		return \wp_remote_post(
+			SettingsFriendlyCaptcha::getEndpointUrl(),
```
iruzevic: this should be in this class and not the settings class just like in other locations

src/Captcha/SettingsCaptchaProvider.php
```diff
@@ -0,0 +1,143 @@
...
+use EightshiftForms\Helpers\SettingsOutputHelpers;
+use EightshiftForms\Settings\SettingGlobalInterface;
+use EightshiftFormsVendor\EightshiftLibs\Services\ServiceInterface;
+
+/**
+ * Wraps the two per-provider settings pages behind a single captcha menu entry
+ * with a provider selector. Routes the admin UI and runtime validity checks
+ * through whichever provider is currently active.
+ */
+class SettingsCaptchaProvider implements SettingGlobalInterface, ServiceInterface
```
iruzevic: the same naming standard as for Google ReCpatcha. Main settings page should be named SettingsCaptcha.php and Google SettingsCaptchaGoogle, etc

src/Captcha/SettingsFriendlyCaptcha.php
```diff
@@ -0,0 +1,230 @@
...
+
+	/**
+	 * Friendly Captcha use EU endpoint key.
+	 */
+	public const SETTINGS_FRIENDLY_CAPTCHA_USE_EU_ENDPOINT_KEY = 'friendly-captcha-use-eu-endpoint';
+
+	/**
+	 * Friendly Captcha API endpoint URLs.
+	 */
+	public const FRIENDLY_CAPTCHA_ENDPOINT_GLOBAL_URL = 'https://global.frcapi.com/api/v2/captcha/siteverify';
```
iruzevic: This should all go to file where you connect to captcha and not here

src/Enqueue/Blocks/EnqueueBlocks.php
```diff
@@ -346,6 +349,22 @@ public function enqueueBlockFrontendScript(): void
 			];
 		}
 
+		// Check if Friendly Captcha data is set and valid.
+		if (\apply_filters(SettingsFriendlyCaptcha::FILTER_SETTINGS_GLOBAL_IS_VALID_NAME, false)) {
+			$output['friendlyCaptcha'] = [
```
iruzevic: as discoused in JS this sould be the same flag for both, is catpcha used bool and then options of that captcha and maybe type to determin the usage later

src/Hooks/FiltersSettingsBuilder.php
```diff
@@ -150,14 +151,13 @@ public function getSettingsFiltersData(): array
 					'title' => \__('Advanced', 'eightshift-forms'),
 				],
 			],
-			SettingsCaptcha::SETTINGS_TYPE_KEY => [
-				'settingsGlobal' => SettingsCaptcha::FILTER_SETTINGS_GLOBAL_NAME,
+			'captcha' => [
```
iruzevic: why did you change this?

src/Labels/Labels.php
```diff
@@ -425,6 +425,9 @@ private function getCaptchaLabels(): array
 			'captchaScoreSpam' => \__('The request was marked as a potential spam request. Please try again.', 'eightshift-forms'),
 			'captchaError' => \__('Spam prevention system encountered an error. Please try again.', 'eightshift-forms'),
 			'captchaSuccess' => \__('Success', 'eightshift-forms'),
+			'friendlyCaptchaBadRequest' => \__('Spam prevention system encountered an error. Friendly Captcha request is invalid or malformed.', 'eightshift-forms'),
```
iruzevic, line 428: Have you tested other possible outputs for the CAPTCHA API to determine if we can provide meaningful messages to the user other than generic ones?

src/Rest/Routes/General/CaptchaValidateRoute.php
```diff
@@ -42,18 +43,35 @@ class CaptchaValidateRoute extends AbstractSimpleFormSubmit
...
+	 * Register the route only when Google reCAPTCHA is the active provider.
+	 *
+	 * Friendly Captcha solves entirely in the browser and has no pre-submission
+	 * verification endpoint — exposing this route for it would misroute tokens.
+	 *
+	 * @return void
+	 */
+	public function register(): void
+	{
+		if (SettingsCaptchaProvider::getActiveProvider() !== SettingsCaptchaProvider::PROVIDER_GOOGLE) {
```
iruzevic: Checking settings in register methods is not the best idea as you can't be sure it will 100% be initialised. And this check is done in the check method so no need for duplicate check

src/Rest/Routes/Integrations/ActiveCampaign/FormSubmitActiveCampaignRoute.php
```diff
@@ -59,15 +59,15 @@ public function __construct(
...
-		CaptchaInterface $captcha,
+		CaptchaInterface $captchaDispatcher,
 		MailerInterface $mailer,
 		EnrichmentInterface $enrichment,
 		ActiveCampaignClientInterface $activeCampaignClient
 	) {
 		$this->security = $security;
 		$this->validator = $validator;
 		$this->labels = $labels;
-		$this->captcha = $captcha;
```
iruzevic, line 70: all this should be refactored when the correct init setup of the captcha classes are done

src/Rest/Routes/AbstractIntegrationFormSubmit.php
```diff
@@ -1273,6 +1273,10 @@ protected function prepareApiParams(WP_REST_Request $request, string $type = sel
 					$output[Config::FD_CAPTCHA] = $value['value'];
 					$output[Config::FD_PARAMS][$key] = $value;
 					break;
+				case UtilsHelper::getStateParam('friendlyCaptcha'):
```
iruzevic: use the same param, no need for duplications

src/Troubleshooting/SettingsFallback.php
```diff
@@ -97,6 +97,12 @@ class SettingsFallback implements ServiceInterface, SettingsFallbackDataInterfac
 	public const SETTINGS_FALLBACK_FLAG_CAPTCHA_SUCCESS = 'captchaSuccess';
 	public const SETTINGS_FALLBACK_FLAG_CAPTCHA_DEBUG_SKIP_CHECK = 'captchaDebugSkipCheck';
 
+	public const SETTINGS_FALLBACK_FLAG_FRIENDLY_CAPTCHA_FEATURE_DISABLED = 'friendlyCaptchaFeatureDisabled';
+	public const SETTINGS_FALLBACK_FLAG_FRIENDLY_CAPTCHA_REQUEST_MISSING_TOKEN = 'friendlyCaptchaRequestMissingToken';
+	public const SETTINGS_FALLBACK_FLAG_FRIENDLY_CAPTCHA_REQUEST_WP_ERROR = 'friendlyCaptchaRequestWpError';
+	public const SETTINGS_FALLBACK_FLAG_FRIENDLY_CAPTCHA_OUTPUT_ERROR = 'friendlyCaptchaOutputError';
+	public const SETTINGS_FALLBACK_FLAG_FRIENDLY_CAPTCHA_SUCCESS = 'friendlyCaptchaSuccess';
```
iruzevic: are these constants ever used in this file for failback?

iruzevic (changes requested):
This PR still needs major refactoring of the architecture.

1. Follow previously created files and settings for examples
2. Don't invent the wheel
3. The setting page looks bad
4. Reuse things you already have.
5. Corvus setting has ability to show different settings based on the selected field

Image

src/Blocks/components/form/assets/friendly-captcha.js
```diff
@@ -0,0 +1,115 @@
...
+	 * Init all actions.
+	 *
+	 * @returns {void}
+	 */
+	init() {
+		if (!this.state.getStateCaptchaIsUsed()) {
+			return;
+		}
+
+		if (this.state.getStateCaptchaType() !== StateEnum.CAPTCHA_TYPE_FRIENDLY) {
```
iruzevic: no need to check the type as you already loaded the class

src/Blocks/components/form/assets/friendly-captcha.js
```diff
@@ -0,0 +1,115 @@
...
+			return;
+		}
+
+		if (!this.state.getStateCaptchaLoadOnInit() && !document.querySelectorAll(this.state.getStateSelector('form', true))?.length) {
```
iruzevic, line 35: the whole point of loading the cpactha on every page i not to limit the load on the sites that have forms!

src/Captcha/FriendlyCaptcha.php
```diff
@@ -0,0 +1,251 @@
...
+				AbstractBaseRoute::R_MSG => $this->labels->getLabel('friendlyCaptchaSuccess'),
+				AbstractBaseRoute::R_DEBUG => [
+					AbstractBaseRoute::R_DEBUG_KEY => SettingsFallback::SETTINGS_FALLBACK_FLAG_CAPTCHA_SUCCESS,
+					AbstractBaseRoute::R_DEBUG => $debug,
+				],
+			];
+		}
+
+		// Auth issues — bad/missing API key. Status 401/403 or specific error codes.
+		if (\in_array($responseCode, [401, 403], true) || \in_array($errorCode, self::ERROR_CODES_AUTH, true)) {
```
iruzevic: you have helpers for isSuccessResponse and isErrorResponse please use them

iruzevic (commented):
a few more comments

iruzevic (approved):

piqusy (approved):

PR #617, opened 2026-04-22
Fix reCAPTCHA Enterprise recoverable-error retry handling

ghost:
# Description

### Fixed

- Fixed reCAPTCHA Enterprise failing hard on recoverable `invalidReason` values. `EXPIRED`, `DUPE`, and `TIMEOUT_OR_DUPLICATE` now trigger a fresh-token client retry alongside the existing `BROWSER_ERROR` retry. This resolves the common failure where candidates on slow connections exceed the 2-minute Enterprise token TTL during large resume uploads on Workable career forms — Google returns `tokenProperties.valid = false, invalidReason = "EXPIRED"` and the plugin was treating it as a dead-end error.
- Fixed the free reCAPTCHA v3 path treating `timeout-or-duplicate` as fatal; it now retries on both `browser-error` and `timeout-or-duplicate` for consistency with the Enterprise path.

### Changed

- Added Google-recommended `userAgent` and `userIpAddress` fields to the reCAPTCHA Enterprise `assessments` request body to improve fraud detection (per [Google's current docs](https://cloud.google.com/recaptcha/docs/create-assessment-website)). The IP is resolved via the existing Cloudflare/CloudFront-aware `Security::getIpAddress()` helper; both fields are omitted from the payload if the source value is empty.
- Surfaced `invalidReason` (Enterprise) and `error-codes` (free v3) at the top of the captcha error debug payload so they are visible in the activity log without having to dig into the raw API response body. The original ticket that prompted this change could not be diagnosed from the log alone for exactly this reason.
- Bumped plugin version to 9.1.5.

Note: this complements #603, which handled the client-side `reCAPTCHA Timeout` JS-layer case. This PR addresses the server-side case where Google returns a response but marks the token invalid with a recoverable reason.

## QA Guide

1. **Happy path:** submit a form using reCAPTCHA Enterprise on a normal connection — it should work as before. Server logs for the `assessments` request should now include `userAgent` and `userIpAddress` in the body.
2. **Expired token recovery (main fix):** on a form that uses reCAPTCHA Enterprise and a large file upload (e.g. a Workable career form with resume), throttle the browser's network in DevTools ("Slow 3G") so the upload exceeds 2 minutes. The form should now automatically regenerate a fresh token and succeed on retry, rather than showing the generic captcha error.
3. **Free v3 regression:** on a site using free reCAPTCHA v3, confirm normal submissions still work.
4. **Activity log diagnostics:** force any invalid-token scenario (e.g. temporarily break the site key) and confirm the new `invalidReason` / `errorCodes` field is visible at the top of the debug object in the activity log.

**Expected result:** transient token-expiry failures recover automatically; normal submissions behave as before; activity log entries for captcha failures are easier to diagnose.

# Screenshots / Videos

# Linked documentation PR

N/A

# Productive ticket

N/A

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Captcha/Captcha.php
+++ b/src/Captcha/Captcha.php
@@ -153,8 +153,7 @@ class Captcha implements CaptchaInterface
 		];
 
 		// Google recommends sending userAgent and userIpAddress to improve detection.
-		// phpcs:ignore WordPress.Security.NonceVerification.Recommended
-		$userAgent = isset($_SERVER['HTTP_USER_AGENT']) ? \sanitize_text_field(\wp_unslash($_SERVER['HTTP_USER_AGENT'])) : '';
+		$userAgent = $this->security->getUserAgent();
 		if ($userAgent !== '') {
 			$event['userAgent'] = $userAgent;
 		}
--- a/src/Captcha/Captcha.php
+++ b/src/Captcha/Captcha.php
@@ -16,6 +16,7 @@ use EightshiftForms\Exception\BadRequestException;
 use EightshiftForms\Helpers\SettingsHelpers;
 use EightshiftForms\Helpers\UtilsHelper;
 use EightshiftForms\Rest\Routes\AbstractBaseRoute;
+use EightshiftForms\Security\SecurityInterface;
 use EightshiftForms\Troubleshooting\SettingsFallback;
 use WP_Error;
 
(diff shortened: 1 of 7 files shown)


src/Captcha/Captcha.php
```diff
@@ -126,6 +146,24 @@ private function onEnterprise(string $token, string $action): array|WP_Error
...
 
+		$event = [
+			'siteKey' => $siteKey,
+			'token' => $token,
+			'expectedAction' => $action,
+		];
+
+		// Google recommends sending userAgent and userIpAddress to improve detection.
+		// phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		$userAgent = isset($_SERVER['HTTP_USER_AGENT']) ? \sanitize_text_field(\wp_unslash($_SERVER['HTTP_USER_AGENT'])) : '';
```
iruzevic: why not move this to the security class as you have the same thing in the abstractBaseRoute so you can use the same thing in the same place?

iruzevic (changes requested):
One small optimisation, everything else looks awesome.

iruzevic (approved):
Excelent, great job

PR #646, opened 2026-07-24
feature/select-dropdown-body-end

iruzevic:
# Description

Adds a new accessibility setting, **"Select dropdown location before body end"**, that renders Choices.js dropdowns (select, country, phone) in a floating wrapper at the end of `<body>` instead of inside the field. This prevents the dropdown from being clipped by ancestors with `overflow: hidden` or a `transform`/`contain` containing block.

### Added

- Added the `SETTINGS_GENERAL_A11Y_DROPDOWN_BEFORE_BODY_END_KEY` setting with a toggle in the general accessibility settings, and exposed it to the frontend via the `formDropdownBeforeBodyEnd` localization output.
- Added `setupSelectDropdownPortal()` in `form.js` that portals the open dropdown to `<body>`, mirrors the outer classes so scoped styles still apply, handles positioning/flip, re-binds Choices.js delegated mouse/keyboard handlers, manages Tab focus, and cleans up on close/destroy.
- Added the `getStateSettingsDropdownBeforeBodyEnd` state getter and the `dropdownBeforeBodyEnd` state key wired through `state.js` / `state-init.js`.

### Fixed

- Fixed active/focus state handlers in `utils.js` and `form.js` throwing when the event target is detached from its field (e.g. the portaled dropdown's search input) by guarding against a missing field.

## QA Guide

Steps for QA to test this feature:

1. Go to **Eightshift Forms → Global Settings → General → Accessibility** and enable **"Select dropdown location before body end"**.
2. On the frontend, open a form with a select, country, or phone field — ideally one placed inside a container with `overflow: hidden` or a CSS `transform` (e.g. a modal, slider, or scroll area).
3. Open the dropdown and verify it renders fully (not clipped), positioned correctly below/above the field.
4. Verify option selection via mouse and keyboard, search filtering, flip-when-no-space-below, closing on outside click, and Tab moving focus back to the field.
5. Disable the setting and confirm the dropdown reverts to rendering under the field as before.

**Expected result:** With the setting on, dropdowns render at the end of `<body>` and are never clipped, with mouse/keyboard/search/focus behaving identically to the default inline dropdown. With it off, behavior is unchanged.

# Screenshots / Videos

# Linked documentation PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 This projects adheres to [Semantic Versioning](https://semver.org/) and [Keep a CHANGELOG](https://keepachangelog.com/).
 
+## [9.11.0]
+
+### Added
+
+- Added a "Select dropdown location before body end" accessibility setting that renders select, country, and phone dropdowns before the `</body>` tag instead of under the field.
+
 ## [9.10.1]
 
 ### Fixed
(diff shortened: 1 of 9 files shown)


goranalkovic-infinum (approved):

piqusy (approved):

piqusy:
This is a backport to Eightshift Forms v9 which is no longer needed, v10 has this implemented

PR #654, opened 2026-09-16
Exempt verified C2PA manifests from PDF embedded-file rejection

iobrado:
# Description

Exempt verified C2PA "Content Credentials" manifests from the PDF scanner's embedded-file rejection, so files exported by Adobe, Canva and OpenAI stop being rejected outright. Off by default, and granted only on a qpdf-expanded body.

### PDF scanner

- Report *which* dangerous keys matched instead of a bare boolean (`getMatchedKeys`)
- Replace the tri-state `?bool` assessment with a `PdfVerdict` enum — `Safe` / `Unsafe` / `Undetermined` — so every call site handles "cannot tell" explicitly
- Exempt a body only when the matched set is exactly `/EmbeddedFile` and/or `/EmbeddedFiles`
- Keep rejecting anything carrying a manifest alongside another dangerous key
- Grant the exemption on qpdf output only. A raw body resolves `2 0 R` by reading top to bottom, a reader resolves it through the xref table, so a body that defines an object twice can show the verifier a manifest while the reader extracts a ZIP. Without qpdf the file is rejected exactly as before
- Keep qpdf's output when it exits with warnings. qpdf warns about a large share of real PDFs — a reconstructed xref, a repaired page tree — while still writing a faithful expansion, and discarding it left compressed object streams unexamined. This one affects every PDF upload, not just C2PA ones
- Extract `PdfTokens` so the scanner and the verifier cannot drift on what a present name token means

### C2PA verification

- Add `C2paManifestVerifier` — inspects embedded payload bytes for a well-formed JUMBF superbox carrying the registered C2PA content-type UUID
- Ignore `/AFRelationship` and `/Subtype`; both are attacker-controlled labels
- Read dictionaries with a balanced-bracket walk rather than regex text matching: `/EF` resolves direct and indirect values alike and rejects any shape it cannot resolve, and `/Length` is read from the masked outer entries only, requiring exactly one match
- Reject a body that defines the same object twice, and bound the stream lookup to the referenced object so an `/EF` pointing at a streamless object cannot borrow an unrelated payload
- Reject `/RF` (Related Files) outright, and backstop it by rejecting any embedded-file stream the `/EF` walk never verified — qpdf keeps an `/RF` target through `--qdf`, and `/Type` is optional so a check keyed on it alone is blind
- Fail closed on object streams, unresolvable refs, filtered streams, unreadable lengths and non-JUMBF payloads
- Structural check only — makes no claim about signature authenticity

### Configuration

- Add `fileSecurityPdfAllowC2pa` filter; the exemption is inactive until a site opts in
- Compare with `=== true` so a truthy non-boolean filter return (`1`, `'yes'`) fails closed
- Add `fileSecurityPdfDangerousKeys` filter to let integrators adjust the key list
- Add `FILE_UPLOAD_PDF_C2PA_MAX_BYTES` (2 MB) to bound how much opaque data the exemption can carry

Opting in — nothing changes until a site adds this, and qpdf must be installed on the host:

```php
\add_filter('es_forms_validation_file_security_pdf_allow_c2pa', '__return_true');
```

The key list is filterable too, if a site needs to tighten or loosen what counts as dangerous:

```php
\add_filter('es_forms_validation_file_security_pdf_dangerous_keys', static function (array $keys): array {
	return \array_merge($keys, ['/GoToR']);
});
```

### Tests

- Add `tests/security/verify-c2pa.php` and `tests/security/verify-pdf-scanner.php`, dependency-free harnesses (project has no PHP test framework)
- Share the JUMBF box builders through `tests/security/c2pa-fixtures.php` instead of four independent copies
- Fixtures: genuine manifest, mislabelled payload, JUMBF-prefixed ZIP, manifest + ordinary attachment, manifest + JavaScript, and a compressed object stream that makes qpdf exit with warnings
- Scanner-level coverage for the decision that matters: the exemption stays off by default, needs qpdf, does not extend to a manifest sitting beside `/JavaScript`, and rejects a shadow object whose xref points at a ZIP
- Regression cases for every bypass closed here, each confirmed to fail against the pre-fix verifier: indirect `/EF`, decoy `/Length`, duplicate object definitions, `/RF` payloads, and embedded-file streams the walk never reached
- Cover two bugs found against real files: bare CR line endings as Adobe writes them, and multi-digit indirect `/Length` object numbers
- Gitignore generated fixture output

### Release

- Bump to 10.5.0 — plugin header, `package.json` and CHANGELOG. `package.json` had drifted to 10.0.0 across earlier releases and is realigned here


src/Validation/FileSecurity/C2paManifestVerifier.php
```diff
@@ -0,0 +1,866 @@
...
+	 * end inside binary stream data, and hid genuine definitions whenever a
+	 * payload happened to carry a `%` byte.
+	 *
+	 * @param string $body qpdf-expanded PDF bytes.
+	 *
+	 * @return array<string, int> `"number generation"` key to byte offset.
+	 */
+	private function mapObjectOffsets(string $body): array
+	{
+		$count = \preg_match_all(
```
piqusy, line 359: [P1] Bound object-offset discovery to parsed object regions. This regex scans opaque stream payloads, so malformed fake `N 0 obj <<` sequences inside a manifest can cause repeated dictionary walks toward the body end and exhaust CPU or memory on an opted-in upload. Add a regression fixture with fake definitions inside an otherwise valid manifest.

src/Validation/FileSecurity/C2paManifestVerifier.php
```diff
@@ -0,0 +1,866 @@
...
+
+		$references = $this->collectEmbeddedFileReferences($objects);
+
+		if ($references === []) {
+			return false;
+		}
+
+		$verified = [];
+
+		foreach ($references as [$number, $generation]) {
```
piqusy, line 109: [P1] Skip already-verified object/generation pairs before resolving and parsing the payload. A PDF can put the same large manifest behind many `/EF` references; `$verified` is only populated after the full walk, so validation cost scales with references rather than unique streams. Add repeated-reference coverage.

tests/security/verify-pdf-scanner.php
```diff
@@ -0,0 +1,242 @@
...
+			printf("SKIP  %s (%s not generated — run ./generate-test-files.sh first)\n", $label, $file);
+			return;
+		}
+
+		$check($label, $scan($path, $filters), $expected);
+	};
+
+	$unsafe = Labels::LABEL_VALIDATION_FILE_PDF_UNSAFE;
+
+	echo "--- dangerous keys, exemption off (the default) ---\n";
```
piqusy, line 121: [P2] Add focused behavioral coverage for the new dangerous-key filter and the 2 MB C2PA boundary. The current harness does not fail if the cap or `fileSecurityPdfDangerousKeys` override is removed or ignored.

piqusy (dismissed):

goranalkovic-infinum (approved):

tests/security/verify-pdf-scanner.php
```diff
@@ -0,0 +1,279 @@
...
+	$checkFixture('/Launch rejects', 'pdf-launch.pdf', [], $unsafe);
+	$checkFixture('/EmbeddedFile rejects', 'pdf-embedded.pdf', [], $unsafe);
+
+	// The exemption must stay off until a site asks for it. Every other check in
+	// this file is worth less than this one.
+	$checkFixture('genuine C2PA manifest still rejects by default', 'pdf-c2pa-valid.pdf', [], $unsafe);
+
+	echo "\n--- exemption on ---\n";
+
+	if ($hasQpdf) {
```
piqusy, line 133: [P2 RESOLVED] qpdf-enabled verification is now complete on 87be868a with qpdf 12.4.1. The genuine-manifest acceptance path and exit-3 warning path pass, along with the qpdf-expanded exemption and shadow-object checks.

piqusy (commented):
P2 validation blocker: I rechecked head `87be868a`. The local runner has no qpdf, and the passing PR checks do not execute the security harnesses or qpdf. The qpdf-expanded exemption, shadow-object xref resolution, and exit-code 3 warning handling therefore remain unverified. Please run both existing security harnesses on a qpdf-enabled runner and provide non-skipped output.

piqusy:
Follow-up verification for head 87be868a: qpdf 12.4.1 installed locally and both security harnesses now pass without qpdf skips.

- C2PA harness: qpdf --qdf round trip, shadow-object, repeated-reference, fake-definition, and size-cap cases pass.
- PDF scanner harness: genuine C2PA acceptance, qpdf exit 3 handling, shadow-object rejection, and dangerous-key filter cases pass.

The prior P2 validation gap is closed. No new code findings in the latest head.

piqusy:
Fresh final checkpoint for head 87be868a: qpdf-backed verification is complete and the checkpoint reviewer returned PROCEED. Both security harnesses pass without skips, including qpdf round-trip acceptance, exit-code 3 handling, shadow-object rejection, amplification regressions, size-cap, and dangerous-key filter cases. No evidenced blockers or new findings remain.

